PR #72, opened 2011-11-01
Make CanCan work with Draper Decorators

momolog:
Make decorators be kind_of? their models class, so that cancan rules apply to them.
This makes CanCan seamlessly work with decorators as it would with the corresponding model instances.

### DIFF
--- a/lib/draper/base.rb
+++ b/lib/draper/base.rb
@@ -146,6 +146,11 @@ module Draper
       @model == (other.respond_to?(:model) ? other.model : other)
     end
 
+    def kind_of?(klass)
+      return true if klass == model.class
+      super
+    end
+
     def respond_to?(method, include_private = false)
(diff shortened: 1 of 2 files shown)


Alexander-Senko:
I consider this to be reverted for the following reasons:

1. It's terribly wrong to pretend being a class without implementing it's interface. It can [break third party code](https://github.com/drapergem/draper/issues/859).
2. Hacking Ruby core methods can lead to unexpected and hard to understand behavior. It should be avoided without a strong reason behind. 
3. I'd prefer compatibility patches to be narrow-scoped and live in there own modules.
4. This [doesn't seem to fix CanCanCan](https://github.com/CanCanCommunity/cancancan/issues/255#issuecomment-146618381) for a long time now. Luckily, there is [`draper-cancancan`](https://github.com/GoodMeasuresLLC/draper-cancancan).

PR #579, opened 2013-10-15
Resolving some nuances with equality

brocktimus:
I noticed some strangeness with how equality works on decorated objects. This is particularly prevalent when dealing with collections of objects and using methods provided by Enumerable. This is a pull request to try and resolve this, at the moment it is just a failing test.

An easy example:
```ruby
model = Model.first
[model, model].uniq.count => 1
[model.decorate, model.decorate].uniq.count => 2
```

Looking into it in more detail it seems a lot of these methods rely on a combination of ```#eql?``` and ```#hash```. Both of these return unexpected results currently. This caused a lot of confusion since ```#group_by``` on a collection wouldn't group, despite many keys being equal. This is particularly a problem when using decorated associations.

This is shown below:
```ruby
irb(main):006:0> model.decorate.eql? model.decorate
=> false
irb(main):007:0> model.decorate.hash
=> -521953688451298283
irb(main):008:0> model.decorate.hash
=> 573320235355283849
```

I "fixed" this in the problem area by adding the following two methods to the decorator which I was using.
```ruby
ModelDecorator < Draper::Decorator
  def eql?(other)
    self == other
  end

  def hash
    object.hash
  end
end
```

I think the ```#eql?``` implementation might be fine. I'm not sure, but I think the ```#hash``` isn't "correct" since it will match the undecorated object.

I'll have more of a look at the code later for a possible fix. How does this sound? Any idea of an appropriate way of generating the hash?

Also any advice where this code would go? I tried to put my ```#eql?``` definition in [Draper::Decoratable::Equality](lib/draper/decoratable/equality.rb) but it still failed the tests (code isn't committed since it didn't help).

### DIFF
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -173,10 +173,17 @@ module Draper
       Draper::Decoratable::Equality.test(object, other)
     end
 
+    # Delegates equality to :== as expected
+    #
+    # @return [Boolean]
     def eql?(other)
       self == other
     end
 
+    # Returns a unique hash for a decorated object based on
+    # the decorator class and the object being decorated.
+    # 
+    # @return [Fixnum]
     def hash
       self.class.hash ^ object.hash
     end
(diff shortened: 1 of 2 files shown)


jcasimir:
Dig it, thanks.

Issue #538, opened 2013-05-09
Relation#decorate() doesn't cache objects returned by joins, ExampleDraper.decorate_collection does


krainboltgreene:
Latest stable Rails, latest stable draper

``` bash
rails new draper_issue --skip-bundle -G
cd draper_issue
echo "gem 'draper'" >> Gemfile
bundle install
rails generate resource Account
rails generate resource Article account:belongs_to
rails generate resource Comment article:belongs_to loved:boolean:index
rake db:create db:migrate
echo "class Account < ActiveRecord::Base\n  has_many :articles\nend" > app/models/account.rb
echo "class Article < ActiveRecord::Base\n  belongs_to :account\n  has_many :comments\n  attr_accessor :loved\nend" > app/models/article.rb
echo "class Comment < ActiveRecord::Base\n  belongs_to :article\n  attr_accessible :loved\nend" > app/models/comment.rb
echo "class AccountDecorator < Draper::Decorator\n  delegate_all\n  decorates_association :articles\nend" > app/decorators/account_decorator.rb
echo "class ArticleDecorator < Draper::Decorator\n  delegate_all\n  decorates_association :comments\nend" > app/decorators/article_decorator.rb
echo "class CommentDecorator < Draper::Decorator\n  delegate_all\nend" > app/decorators/comment_decorator.rb
rails console
```

Which makes these:

``` ruby
class Account < ActiveRecord::Base
  has_many :articles
end

class Article < ActiveRecord::Base
  belongs :account
  has_many :comments

  attr_accessor :loved
end

class Comment < ActiveRecord::Base
  belongs_to :article

  attr_accessible :loved
end

class AccountDecorator < Draper::Decorator
  delegate_all
  decorates_association :articles

end

class ArticleDecorator < Draper::Decorator
  delegate_all
  decorates_association :comments

end

class CommentDecorator < Draper::Decorator
  delegate_all

end
```

Now we seed the database:

``` ruby
account = Account.create
article = account.articles.create
article.comments.create
article.comments.create loved: true
article.comments.create

article = account.articles.create
article.comments.create loved: true
article.comments.create

account = Account.create
article = account.articles.create
article.comments.create
article.comments.create loved: true
article.comments.create

article = account.articles.create
article.comments.create loved: true
article.comments.create
```

Now we show our problem:

``` ruby
# Clear previous bindings
reload!

# Get all accounts, join the article and comments
accounts = Account.joins(articles: :comments).uniq

[...]

PR #588, opened 2013-12-01
Work in progress fix for #538

brocktimus:
More details in #538. This is just extracted from a monkey patch I've used briefly in an application as discussed there.

This obviously needs testing but I'm unsure best / most appropriate way to go about it. It requires a lot more "stuff" than currently exists in the test suite (I think...?).

I also thought it should go into its own name space or something for other optimizations? I'm guessing something similar could be written for other AR relations and potentially Mongoid as well.

Any thoughts / advice appreciated!

### DIFF
--- a/lib/draper.rb
+++ b/lib/draper.rb
@@ -51,6 +51,14 @@ module Draper
     end
   end
 
+  def self.setup_active_record_collection_proxy
+    ActiveRecord::Associations::CollectionProxy.class_eval do
+      def decorate(options = {})
+        @association.klass.decorator_class.decorate_collection(@association.target, options.reverse_merge(with: nil))
+      end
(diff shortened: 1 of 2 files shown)


jcasimir:
Seems really interesting but we need test support / documentation.

brocktimus:
Based on #538 is this one also invalid if ```#decorate``` is being removed?

Otherwise I can look at it more just might need a bit of advice in best way of testing it. Most of the current testing stuff didn't seem to be dependent on having a full blown AR setup which this seems to start to require.

Alexander-Senko:
Just came across this PR accidentally. It'd better have a more descriptive name to be discovered.

So, it's to be closed in favor of #932.

PR #596, opened 2014-01-08
Bring back compatibility with ActiveModel::Serializers

stevenharman:
The current AMS re-write (which will debut as v0.9.0) has changed how serializers are resolved, and removed the `ArraySerializerSupport` module. This change will ensure Draper properly integrates with both AMS pre-0.9.0, and the future.

I went with the obvious approach, but another option would be to instead test the current loaded version of `active_model_serializers` instead. So, sub out those `if defined?(ActiveModel::ArraySerializerSupport)` checks for something like `Gem.loaded_specs['active_model_serializers'].version.release < Gem::Version.new('0.9')`.

Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,6 @@ platforms :jruby do
   gem "activerecord-jdbcsqlite3-adapter", ">= 1.3.0.beta2"
 end
 
-version = ENV["RAILS_VERSION"] || "4.0"
+version = ENV["RAILS_VERSION"] || "4.1"
 
 eval_gemfile File.expand_path("../gemfiles/#{version}.gemfile", __FILE__)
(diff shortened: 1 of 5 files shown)


stevenharman:
Well... I've tried. The code presented here does work for older AMS 0.8.x and the 0.9.x branch. I have no idea about the unstable and actively evolving 0.10x. (a.k.a. `master`) branch. That said, Draper itself is in a pretty bad state w/r/t its tests.

The suite can't currently pass on Rails 4.1 due to changes in MiniTest and how the generators work - I've tried. The tests are also broke on Ruby 2.2 due to a change (I believe) in how constant lookup happens. I tried to get that fixed too, but it's become a mangled mess.

I would like to get this change merged in, despite the failing CI builds, and then use other PRs to drop Rails 3.x support and get proper Rails 4.0-4.2 support. And perhaps another to get Ruby 2.2 support.

aceofspades:
Personally I have abandoned AMS for https://github.com/ismasan/oat

PR #611, opened 2014-04-01
Decorate structs as singular objects (and not as collections)

killpack:
Since structs respond to `.first`, Draper thinks they're collections and
always tries to wrap them in CollectionDecorators, which is surprising!

This is similar to the issue and fix in [#259].

### DIFF
--- a/lib/draper/factory.rb
+++ b/lib/draper/factory.rb
@@ -75,7 +75,7 @@ module Draper
       end
 
       def collection?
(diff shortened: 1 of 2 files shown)


jcasimir:
Dig it, thanks.

mkamensky:
I have a model that responds to :first, which leads to the same problem. Perhaps it is possible to add an explicit option to decorates_association ? Thanks

PR #631, opened 2014-07-01
Update README, extra kaminari methods delegation.

renatomoya:
- In order to use kaminari's helper method `page_entries_info`, you
    need to delegate `:entry_name`, `:total_count`, `:offset_value` and
    `:last_page?`.


jcasimir:
Thank you!

PR #656, opened 2015-03-02
required v0.8.0 for active_model_serializers in gemspec

jgu2160:
This makes all tests pass on rails 4+.


mikedao:
This does not make all the tests pass. It is one step to get tests running in the dummy app.

PR #659, opened 2015-03-09
Remove draper.setup_active_model_serializers

### LINKED ISSUE #644: draper doesn't work with ActiveModelSerializer 0.9.3
I caught the following error:

```
/Users/tricknotes/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/draper-1.4.0/lib/draper/railtie.rb:48:in `block (2 levels) in <class:Railtie>': uninitialized constant ActiveModel::ArraySerializerSupport (NameError)
```

see: https://github.com/drapergem/draper/blob/v1.4.0/lib/draper/railtie.rb#L48

Work versions:
* AMS 0.8.x: It has `ActiveModel::ArraySerializerSupport`
* AMS 0.9.0: It has no `ActiveModel::ArraySerializerSupport`. But it doesn't run load hooks - so it works.

Doesn't work version:
* AMS 0.9.1: It has no `ActiveModel::ArraySerializerSupport`. And run load hooks - so it doesn't work.

m5rk:
Fixes #644


jcasimir:
Thanks, we should be able to check this out in the next few days.

PR #610, opened 2014-03-29
Activerecord relations

donaldpiret:
Better handling of activerecord relations. One can now call the decorator anywhere in the scope chain, and keep your relation decorated. Once you call methods that access individual records or arrays of records it automatically converts your result to a collection decorator or a regular decorator.

Would love to get some feedback on this.

PR #662, opened 2015-03-28
Add better support for activerecord relations and mongoid criteria

donaldpiret:
Cleaned up implementation of #610 with Mongoid support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ vendor/bundle
 *.DS_Store
 spec/dummy/log/*
 spec/dummy/db/*.sqlite3
+.idea
--- a/lib/draper.rb
+++ b/lib/draper.rb
@@ -23,6 +23,7 @@ require 'draper/helper_support'
 require 'draper/view_context'
 require 'draper/collection_decorator'
 require 'draper/undecorate'
+require 'draper/relation_decorator'
 require 'draper/decorates_assigned'
 require 'draper/railtie' if defined?(Rails)
 
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -49,15 +49,15 @@ module Draper
 
     module ClassMethods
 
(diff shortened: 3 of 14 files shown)


mikedao:
I think a valuable question here is if we want to explicitly continue support for Mongoid. 

What does everyone think about this? I think a good discussion about it either way would be valuable, arguments for and arguments against.

colinpetruno:
I was working with your branch and it works well but I noticed two small issues.

1. Associations not defined in decorates_association get returned as a RelationDecorator when a scope is applied. It creates an error like ```is not an ActiveModel-compatible object. It must implement :to_partial_path.```
image

2. RelationDecorator doesn't work with ```render collection: Model.decorated_associaton```. It won't error but it doesn't render any of the partials so it seems unable to find the models in RelationDecorator.

donaldpiret:
Hi @colinpetruno, awesome, thanks for reporting these. I'll try to get issue 2 resolved over the next few days or this weekend at the latest.

saneshark:
:+1:

saneshark:
README documentation needed before merge?

donaldpiret:
Hi @saneshark, not sure I understand your question regarding the delegate_all. If your decorator implements delegate_all, the decorated model objects returned by accessing a member of the relation will respect this as well.
The point of distinguishing it as a relations rather than turning it into a collection is to enable further chaining of scopes down the line, so your .decorate call can be called at any time, not necessarily at the end.
It also supports pagination out of the box by delegating the pagination methods to the relation directly and handling the decoration of the returned model instances.

For example you can now do:

User.decorate.page(2).per(1).first and this will return a decorated user instance.

I'm happy to write up documentation for this if there is a real intention of getting this branch merged. Let me know if this is the case and I'll write up some proper docs.

Alexander-Senko:
I'm ready to review and push it forward if one brings it up to date.

PR #665, opened 2015-04-19
Don't do runtime delegation, delegate explicitly

janko-m:
If someone accidentally calls `super` instead of `object` inside a decorator, all hell breaks loose.

``` rb
require "draper"
require "ostruct"

class PostDecorator < Draper::Decorator
  delegate_all

  def title
    super and "overriden title"
  end
end

post = OpenStruct.new(title: "title")
decorated_post = PostDecorator.new(post)

puts decorated_post.title #=> "overriden title"
puts decorated_post.title #=> "title"
```

This patch makes both `super` and `object` work. I'm not sure why runtime delegation was added in the first place, but it seems very wrong to add methods on the fly.

### DIFF
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -596,12 +596,16 @@ module Draper
           expect(decorator.hello_world).to be :delegated
         end
 
-        it "adds delegated methods to the decorator when they are used" do
-          decorator = Decorator.new(double(hello_world: :delegated))
+        it "allows calling `super`" do
+          decorator_class = Class.new(Decorator) do
+            def hello_world
+              super and "overriden hello world"
+            end
+          end
+          decorator = Decorator.new(double(hello_world: "hello world"))
 
-          expect(decorator.methods).not_to include :hello_world
-          decorator.hello_world
-          expect(decorator.methods).to include :hello_world
+          expect(decorator.hello_world).to eq "overriden hello world"
+          expect(decorator.hello_world).to eq "overriden hello world"
         end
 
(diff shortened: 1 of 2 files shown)


janko-m:
Strange, this is exactly the failure that happens without my change. With I run the tests locally with `rake spec` they work.

bolshakov:
Recently I encountered with this issue, and it was debugging hell!

spec/draper/decorator_spec.rb
```diff
@@ -596,12 +596,16 @@ module Draper
...
 
-        it "adds delegated methods to the decorator when they are used" do
-          decorator = Decorator.new(double(hello_world: :delegated))
+        it "allows calling `super`" do
+          decorator_class = Class.new(Decorator) do
+            def hello_world
+              super and "overriden hello world"
+            end
+          end
+          decorator = Decorator.new(double(hello_world: "hello world"))
```
bolshakov: You should use defined `decorator_class` here:

```ruby
decorator = decorator_class.new(double(hello_world: "hello world"))
```

After this change tests pass

bolshakov:
@janko-m Thank you for the fix. Would you mind to update PR?

janko-m:
I'm not using Draper anymore (since I'm not using Rails/ActiveRecord anymore), feel free to cherry-pick my commit and make the necessary change :)

codebycliff:
@janko-m Thanks for this. I have opened a new pull request with the conflicts resolved and the test fixed [here](https://github.com/drapergem/draper/pull/789). I'm going to close this one now.

Issue #663, opened 2015-04-02
ActiveJob Integration


hayesgm:
Draper decorators do not currently implement [globalid](https://github.com/rails/globalid).  The fix is just to call in an initializer:

``` ruby
Draper::Decorator.send(:include, GlobalID::Identification) if defined?(GlobalID)
```

As `rails` includes `globalid` by default, would we want to add this line to the repo or put it in the documentation?

Without this fix, calling `UserMailer.welcome(user.decorate).deliver_later` will raise `ActiveJob:: SerializationError`, due to [this code](https://github.com/rails/rails/blob/31085a5cd47894b6fa397ca22fe8aa552d339ce8/activejob/lib/active_job/arguments.rb#L51).

mikedao:
Modifying the documentation allows for an extremely quick fix. What are your arguments for/against?

EtienneDepaulis:
This fix might be misleading. It works great to to silence the `ActiveJob:: SerializationError` but as it retrieves the object from DB again later, the object is not decorated anymore, thus preventing from using added methods in the emails.

codebycliff:
While this might be misleading, you should still be able to pass a decorated object to a background job without it blowing up. Ideally, a decorated object should behave as if it were the object it is decorating. Here's what I'm proposing:

1. Add: `include GlobalID::Identification if defined? GlobalID` to `Draper::Decorator` so that it doesn't blow up.
2. Add some documentation to the README saying: "While you can pass a decorated object to background jobs, the object won't be decorated when it is deserialized."

Is anyone opposed to this solution?

codebycliff:
Closing this as I believe it is resolved. If someone figures out how to deserialize the record as a _decorated_ object, feel free to open a pull request.

PR #669, opened 2015-05-05
Add note on Active Job integration for #663

hayesgm:
Add a note about Active Job integration which causes Rails serializations issue when passing Draper decorated objects to Rails background mailers.  This patch satisfies issue #663.

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,6 +573,14 @@ ORM](https://github.com/drapergem/draper/wiki/Using-other-ORMs) (including
 versions of Mongoid prior to 3.0), or want to decorate plain old Ruby objects,
 you can include this module manually.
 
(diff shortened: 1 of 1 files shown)


seanlinsley:
Seems like this should be automatically done by Draper itself.

codebycliff:
I agree that draper should take care of this. However, there is concern about this being misleading as the object won't be decorated when it is deserialized (https://github.com/drapergem/draper/issues/663). I'm proposing we solve this via code AND documentation. I'm awaiting a response on that issue to move forward.

codebycliff:
Closing this in favor of new pull request: [Feature: ActiveJob Integration and Documentation](https://github.com/drapergem/draper/pull/817).

PR #673, opened 2015-05-15
Support AMS 0.10.x

ryansch:
I haven't added specs as the top level stuff doesn't have access to `Rails` and I didn't have time right now to figure out the `spec/dummy` stuff.


seanlinsley:
I'm going to close this for now. If you add specs and documentation, we can reopen it.

PR #674, opened 2015-05-21
Finish updating specs to RSpec3 syntax

mcasper:


lucianosousa:
:+1:

seanlinsley:
Thanks for the PR, but I'm going to keep #746 open instead since it's more recent, therefore more likely to be accurate against the current master.

PR #675, opened 2015-05-24
rspec-rails ~> 3.2 uses rails_helper fixes #668

### LINKED ISSUE #668: Generator needs update for Rspec ~> 3.2
I have a User model that I want to decorate, so I started by generating the decorator:
`rails g decorator User`

This created two files:
`/app/decorators/user_decorator.rb`
`/spec/decorators/user_decorator_spec.rb`

user_decorator_spec.rb is populated with this code:

``` ruby
require 'spec_helper'

describe UserDecorator do
end
```

Recent Rspec uses `require 'rails_helper'`. Since I overlooked this difference, the errors had me going in circles for a couple hours.

Perhaps the generator can be smart about this recent development and modify the require line accordingly based on rspec version?

krainboltgreene:
As of rspec-rails 3.2 they generate two files:

  - spec/spec_helper.rb
  - spec/rails_helper.rb

This allows the user to either do pure ruby unit tests or include the entire rails environment. Draper really only works in a rails application if it has the rails environment. This matches ~> 3.2 and uses the rails_helper instead of spec_helper.

### DIFF
--- a/lib/generators/rspec/templates/decorator_spec.rb
+++ b/lib/generators/rspec/templates/decorator_spec.rb
@@ -4,6 +4,6 @@
   require 'spec_helper'
 <% end %>
 
-describe <%= class_name %>Decorator do
+describe <%= class_name %>Decorator, type: :decorator do
   
 end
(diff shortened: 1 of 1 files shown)


seanlinsley:
As of #752 we're dropping support for older versions of Rails, so this extra conditional logic isn't necessary. Thanks for the PR, regardless.

krainboltgreene:
Thanks for the consideration!

PR #677, opened 2015-06-01
update gem version on readme

lucianosousa:
related to #667

PR #682, opened 2015-07-01
Update version in installation section of README

mwear:
The installation section of the README currently instructs users to add `gem 'draper', '~> 1.3'` to their Gemfile. The current version of Draper is 2.1 and it might be better to encourage users to use it.


mwear:
Closing as this appears to be a dup of #677 and it failed CI (somehow).

PR #685, opened 2015-07-09
Change old info about pagination to the new one

MaximAbramchuck:


codebycliff:
Can explain what the "new one" is or provide a description?

codebycliff:
While the documentation change you made makes sense (from what I can understand), I'd rather see this get built into Draper automatically. For example, automatically adding the delegation if the `Kaminari` constant is defined or something. This would provide a more friction free experience. 

I could make a case for adding this to the documentation until we get that functionality added. However, the documentation you provided isn't entirely complete. The README goes on to mention the `PaginatingDecorator` reference that you are removing. If you wanted to update all those references, I'd be happy to merge this if we haven't added the functionality by then.

PR #693, opened 2015-09-09
Update Bundler version specifier in installation instructions for 2.x

jimryan:


lucianosousa:
#677 is doing it

jimryan:
Missed that - but I should note that #677 locks down the version specifier to 2.1.x (~> 2.1.0) , whereas this PR locks it to >= 2.1 and < 3.0 (~> 2.1), which aligns with what the README instructed for 1.x (~> 1.3).

lucianosousa:
I was just following the latest README examples.

Anyway,
looks like there is no maintenance here anymore.

jimryan:
Isn't this the latest README example:

```ruby
  gem 'draper', '~> 1.3'
```

Maybe I'm missing/misunderstanding something, but ~> 2.1.0 is not the same as ~> 2.1, the former would allow >= 2.1.0 and < 2.1, whereas the latter would allow >= 2.1 and < 3.0

lucianosousa:
You're right man. ;)

Makes totally sense. Agree with this PR.

lucianosousa:
Consider remove the infos about the update from version 0.x

Doesn't  make sense anymore, in my opinion. ;)

jimryan:
I agree - I think that should probably be its own PR though.. maybe

seanlinsley:
This has been fixed by not specifying a version in the readme. Thanks for the PR though!

PR #692, opened 2015-08-27
Rspec3 Update - Spec Syntax, Generators, and Monkey-Patching

baberthal:
## Update to latest RSpec (3.3)

As of rspec-rails 3.3 generates two files:
- spec/spec_helper.rb
- spec/rails_helper.rb

This allows the user to either do pure ruby unit tests or include the entire
rails environment. Draper really only works in a rails application if it has the
rails environment. This matches ~> 3.3 and uses the rails_helper instead of
spec_helper for the dummy app specs, but the standard spec_helper for
the draper specs, so the entire suite doesn't depend on loading rails.

Fix Spec spec generator because RSpec no longer prefers
monkey-patching. Refactor PR #675 to be more concise re RSpec Version
check, and to use RSpec::Core::Version rather than
RSpec::Rails::Version, in case the latter does not exist.
## Updated all specs to RSpec 3.0 syntax

Change specs to use synax `expect(OBJECT).to EXPECTATION` rather that
`OBJECT.should`. Changed all `OBJECT.stub(METHOD)` to `allow(OBJECT).to
receive(:method)`

Change one-liners to use syntax: `it { is_expected.to XXX }` rather than
`it { should }`.

Merge PR #674 from @mcasper and fix all trivial conflicts (i.e
`allow(OBJ).to receive(:message)` vs `allow(OBJ).to receive_messages`)
## Gitignore

Ignore .ruby_version, .ruby_gemset, any rspec_output, and any generated binstubs.
## Spec Helper

ActiveRecord::Migration.maintain_test_schema! is undefined in Rails
4.0, so I made that method conditional upon ENV['RAILS_VERSION'].

Explicitly disable monkey_patching, in addition to some trivial edits related to the RSpec upgrade.
## Notes

All the specs pass, although one of the generator specs throws an uninitialized constant RSpec::Support::Differ error. This appears to happen both on travis on on my local machine. I believe it is related to https://github.com/alexrothenberg/ammeter/issues/48, but I can't be sure. Let me know what you guys think and I'll see what I can do to fix it.

PR #695, opened 2015-09-10
Improve spec generator template

zavan:
Newer RSpec versions requires that you call the first describe block on the RSpec class.

I also added the type of the spec and a "pending" message, to make it similar to the default rspec generators.

I kept `spec_helper` instead of `rails_helper` because there's no need to load Rails to test the decorator in isolation, since we can test them with POROs and/or stubbing methods.


zavan:
Didn't see #692 before, sorry, closing this.

PR #696, opened 2015-09-15
Updating generator to work for Rails 4

SpencerCDixon:
When I used the generator I noticed it broke my test suite because it used spec_helper instead of rails_helper


lucianosousa:
#675 has a better improvement for this. (using the type)

seanlinsley:
Thanks!

PR #707, opened 2015-11-13
Optimization: Optimize CollectionDecorator Enumerable behavior

clemens:
Since draper's collection decorator delegates all methods including `each` to its `decorated_collection`, calling `each` will result in two iterations: one setting up the `decorated_collection` via `map` and one doing the iteration the user wants to do. This behavior has negative performance implications especially for large collections.

This PR changes the behavior by explicitly overriding `each` to decorate and store the item in the `@decorated_collection` instance variable and then yield it. Also it removes the delegation of `Enumerable`'s instance methods to the `decorated_collection`. It was unnecessary in the first place because `CollectionDecorator` includes `Enumerable` itself and this way it can leverage the optimized `each` method.

The PR is intended as a basis for discussion. While I assume this covers the most usual use cases, users might be surprised to not see effects of this optimization on array-specific methods such as `each_index`, `map!` and the like.

Happy to hear your thoughts on this!


clemens:
Any comments?

syguer:
`decorated_collection` looks already be cached in instance variable.
https://github.com/drapergem/draper/blob/master/lib/draper/collection_decorator.rb#L41-L43

Could you show us any benchmark?

codebycliff:
@syguer Even though the instance variable is cached, there are still two iterations the first time `each` is called:

1. When calling `map` in `decorated_collection`.
2. When calling `each` on the now decorated collection.

Patch seems to make sense to me. I took some benchmarks (see below) and there seems to be a noticeable improvement in performance. In my benchmarks, I was iterating over a collection of 10,000 records. Nobody should be iterating over that large of a collection in view, but I still think the change makes sense. It has tests and documentation, just needs the conflicts resolved. 

@clemens Sorry we are just now getting back to you. Would you be interested in fixing the conflicts on this?

## Benchmarks

### Non-decorated collection
image

## Decorated collection pre-patch 
image

## Decorated collection with patch
image

codebycliff:
Closing and re-opening to reset the travis build. It's still stuck on old build configurations from when the pull request was originally opened.

clemens:
Fixed the issues. Let me know if I can do anything else. :)

codebycliff:
@clemens Thanks! I don't think I need anything else unless there you want to discuss it more. Did you have any concerns about this change? It seems like it can only help things, but I might be missing something.

clemens:
I don't see any. I've monkeypatched the fix from this PR into a different project before creating the PR and it's been smooth sailing for 1 1/2 years.

codebycliff:
Awesome. I'm going to go ahead and push this through. Thanks for the patch!

PR #708, opened 2015-11-24
Fix bug in respond_to?

TylerRick:
When you call an automatically delegated method on a decorator object, it causes the new method to
remain permanently defined on the decorator class, which causes subsequent calls to `respond_to?` to always return `true` even when the current delegated-to object does not actually respond to that
message.

For example:
```
class A
  def method_on_a
  end
end
class B
end

decorator = Decorator.new(A.new)
decorator.respond_to?(:method_on_a) # => true # as expected
decorator.method_on_a
decorator = Decorator.new(B.new)
decorator.respond_to?(:method_on_a) # => true # incorrect!!
```

This can be very confusing and frustrating if your application uses the same decorator for several different decoratable classes and depends on `respond_to?` to accurately reflect what the underlying decoratable object actually responds to.

I don't yet fully understand what causes this behavior, but it's something to do with the automatic delegation. I added a new test for this and was able to get it passing by adding a `remove_method` after the `delegate` and `send`. This is probably not the best solution, and may have unintended side effects, but I'm currently at a loss how else to solve this.

### DIFF
--- a/lib/draper/automatic_delegation.rb
+++ b/lib/draper/automatic_delegation.rb
@@ -7,7 +7,11 @@ module Draper
       return super unless delegatable?(method)
 
       self.class.delegate method
-      send(method, *args, &block)
+      send(method, *args, &block).tap do
+        self.class.class_eval do
+          remove_method method
+        end
+      end
     end
 
     # Checks if the decorator responds to an instance method, or is able to
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -596,14 +596,6 @@ module Draper
           expect(decorator.hello_world).to be :delegated
         end
 
-        it "adds delegated methods to the decorator when they are used" do
-          decorator = Decorator.new(double(hello_world: :delegated))
-
-          expect(decorator.methods).not_to include :hello_world
-          decorator.hello_world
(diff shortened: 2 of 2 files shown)


TylerRick:
FYI, I ran into a different strange bug in my app when I tried using this branch:

I started getting a `undefined method '_for_user' for #<Array:0x0000000bfa1238>:` error in one of my templates where I had `report.things._for_user(user)`. The `_for_user` method comes from the `ThingsExtensions` module that gets extended into the collection array by this:
```
has_many :things, extend: ThingsExtensions
```

I cannot for the life of me figure out how the `remove_method` would cause this, but commenting that out resolved it, so there you go. I even tried adding a debugger that would stop if the `method` being called was `_for_user` but it never got triggered, so I don't even think that method was going through any decorator.

I ended up just changing my code to use `decorated_thing.source.respond_to?` instead of calling `respond_to?` on the `decorated_thing` directly. Sad, but no other choice at the moment.

sylvain-8422:
I have submitted a simple fix to your problem. #743 You might be interested in it. :-)

We had the same problems here. It was unpleasant to debug! ;-)

seanlinsley:
#743 has been merged, which should fix your issue @TylerRick

PR #710, opened 2015-12-01
converted specs to rspec 3 style

bmalets:
Replace `should` with `expect` in all specs


alpaca-tc:
:+1:

seanlinsley:
Thanks for the PR, but I'm going to keep #746 open instead since it's more recent, therefore more likely to be accurate against the current master.

PR #714, opened 2016-01-18
Fix typo in README

spajic:


seanlinsley:
Thanks for the PR, but this has already been fixed by #711.

PR #717, opened 2016-01-29
Allow to call #super inside decorator

bolshakov:
``` ruby
class Post
  def title
    "calling super from A to Z"
  end
end

class PostDecorator < Draper::Decorator
  delegate_all

  def title
    super.upcase
  end
end

post = PostDecorator.new(Post.new)
post.title #=> "CALLING SUPER FROM A TO Z"
post.title #=> "calling super from A to Z"
```

This commit fixes the issue, so you can safely call super inside overriden method.

### DIFF
--- a/lib/draper/automatic_delegation.rb
+++ b/lib/draper/automatic_delegation.rb
@@ -6,8 +6,12 @@ module Draper
     def method_missing(method, *args, &block)
       return super unless delegatable?(method)
 
-      self.class.delegate method
-      send(method, *args, &block)
+      if self.class.method_defined?(method)
+        object.send(method, *args, &block)
+      else
+        self.class.delegate(method)
+        send(method, *args, &block)
+      end
     end
 
     # Checks if the decorator responds to an instance method, or is able to
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -640,6 +640,17 @@ module Draper
           expect{decorator.hello_world}.to raise_error NoMethodError
           expect(decorator.methods).not_to include :hello_world
         end
+
+        it "allows calling `super`" do
+          decorator = Class.new(Decorator) do
+            def hello_world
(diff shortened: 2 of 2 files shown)


lib/draper/automatic_delegation.rb
```diff
@@ -6,8 +6,12 @@ module AutomaticDelegation
     def method_missing(method, *args, &block)
       return super unless delegatable?(method)
 
-      self.class.delegate method
-      send(method, *args, &block)
+      if self.class.method_defined?(method)
+        object.send(method, *args, &block)
```
bolshakov: Another possible option is to check if method defined and fail with explicit error message with instructions how to fix issue.

spec/draper/decorator_spec.rb
```diff
@@ -640,6 +640,17 @@ module Draper
...
+
+        it "allows calling `super`" do
+          decorator = Class.new(Decorator) do
+            def hello_world
+              super and "overriden hello world"
+            end
+          end.new(double(hello_world: "hello world"))
+
+          expect(decorator.hello_world).to eq "overriden hello world"
+          expect(decorator.hello_world).to eq "overriden hello world"
```
syguer: unnecessary line?

syguer:
umm... I am against this opinion.
* `#super` calls same name method implemented on "superclass", and almost superclass of decorator is `Draper::Decorator`（or `ApplicationDecorator < Draper::Decorator`). So this behavior looks to confuse many people
* It is easy to do same thing
```ruby
class Post
  def title
    "calling super from A to Z"
  end
end

class PostDecorator < Draper::Decorator
  delegate_all

  def title
    object.title.upcase
  end
end
```

codebycliff:
This trying to solve the same thing as https://github.com/drapergem/draper/pull/789, but not addressing the root of the issue (delegation). I'm going to close this in favor of the pull request mentioned.

PR #719, opened 2016-02-04
Fix version in README How to Install section

vimutter:
It was 1.3 for a long time. Now its 2.1 time.


seanlinsley:
Thanks, this has been fixed by removing the version specification. That way it's never out of date!

PR #720, opened 2016-02-05
Only Draper::Decorator types should delegate === to object.

jrafanie:
Draper::Decorater provides the object method and we should delegate === to this
method only for these types.

Other classes could respond_to?(:object) so using is_a? clarifies not just any
class implementing object but only classes that are a Draper::Decorator.

Besides this, another major reason is is_a? is much faster than respond_to?
in ActiveRecord models because of the number of methods AR provides and the
layers of delegation. There is a cache specifically for respond_to? [1] because
it's so slow.

Even with this cache, some of our pages are 10% faster by using the is_a? change
from this patch.

The original implementation [2] used the method source, which was later renamed
to object.  The tests seem to indicate the intention was to always have
this === delegation to source/object for decorated objects... which is what
Drapper::Decorator types are.

Test changes:
* verify decorated instances of Product and instances derived from
decorated instances of Product are === Product.
* verify unrelated decorated instances and non-decorators that
happen to implement object are not === Product.

[1] https://github.com/rails/rails/blob/4-2-stable/activemodel/lib/active_model/attribute_methods.rb#L343-L351
[2] ca116154f37f01faf4d04dcb916b1bca07c64a74

### DIFF
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -86,7 +86,7 @@ module Draper
       #
       # @return [Boolean]
       def ===(other)
-        super || (other.respond_to?(:object) && super(other.object))
+        super || (other.is_a?(Draper::Decorator) && super(other.object))
       end
 
     end
--- a/spec/draper/decoratable_spec.rb
+++ b/spec/draper/decoratable_spec.rb
@@ -109,19 +109,25 @@ module Draper
       end
 
       it "is true for a decorated instance" do
-        decorator = double(object: Product.new)
+        decorator = Product.new.decorate
 
         expect(Product === decorator).to be_truthy
       end
 
       it "is true for a decorated derived instance" do
-        decorator = double(object: Class.new(Product).new)
+        decorator = Class.new(Product).new.decorate
 
         expect(Product === decorator).to be_truthy
       end
 
       it "is false for a decorated unrelated instance" do
-        decorator = double(object: Model.new)
+        decorator = Other.new.decorate
+
+        expect(Product === decorator).to be_falsey
+      end
+
+      it "is false for a non-decorator which happens to respond to object" do
+        decorator = double(object: Product.new)
 
         expect(Product === decorator).to be_falsey
       end


jrafanie:
Here's an example rubyprof screenshot showing that respond_to? is not exactly fast when run many thousands of times:
screen shot 2016-02-05 at 6 04 46 pm

jrafanie:
Note, an unrelated test, spec/generators/controller/controller_generator_spec.rb, fails locally when I run `RAILS_VERSION=4.2 be rake`.  

I'm not sure how to fix that:

```
~/.gem/ruby/2.2.4/gems/rack-1.6.4/lib/rack/builder.rb:146:in `to_app': missing run or map statement (RuntimeError)
  from ~/.gem/ruby/2.2.4/gems/rack-1.6.4/lib/rack/builder.rb:160:in `block in generate_map'
  from ~/.gem/ruby/2.2.4/gems/rack-1.6.4/lib/rack/builder.rb:160:in `each'
  from ~/.gem/ruby/2.2.4/gems/rack-1.6.4/lib/rack/builder.rb:160:in `generate_map'
  from ~/.gem/ruby/2.2.4/gems/rack-1.6.4/lib/rack/builder.rb:145:in `to_app'
  from ~/.gem/ruby/2.2.4/gems/capybara-2.6.2/lib/capybara/rails.rb:12:in `<top (required)>'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `require'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `block in require'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:240:in `load_dependency'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `require'
  from ~/.gem/ruby/2.2.4/gems/rspec-rails-3.4.2/lib/rspec/rails/vendor/capybara.rb:7:in `<top (required)>'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `require'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `block in require'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:240:in `load_dependency'
  from ~/.gem/ruby/2.2.4/gems/activesupport-4.2.5.1/lib/active_support/dependencies.rb:274:in `require'
  from ~/.gem/ruby/2.2.4/gems/rspec-rails-3.4.2/lib/rspec/rails.rb:13:in `<top (required)>'
  from ~/.gem/ruby/2.2.4/gems/ammeter-1.1.3/lib/ammeter/init.rb:2:in `require'
  from ~/.gem/ruby/2.2.4/gems/ammeter-1.1.3/lib/ammeter/init.rb:2:in `<top (required)>'
  from ~/Code/draper/spec/generators/controller/controller_generator_spec.rb:3:in `require'
```

codebycliff:
@jrafanie This makes perfect since and seems like a good change. Instead of just deleting all the tests, do you think we could just get them to pass? I think the first two could be changed to have: (1) `decorator = Product.new.decorate` and (2) `decorator = Class.new(Product).new.decorate`. That makes the test suite pass. I'm not sure about the third test... It passes, but I don't know if that's still a valid test or not. Let me know what you think and then we can get this puppy merged.

jrafanie:
thanks @codebycliff, I'll try to get those tests fixed and update the PR

jrafanie:
@codebycliff updated, let me know what you think

codebycliff:
Looks good! Do you think this pull request solves the issue that https://github.com/drapergem/draper/pull/774 is trying to solve? I believe it would, but wanted to get a second opinion..

codebycliff:
@jrafanie Actually, I have a couple more things:

1. After thinking on it, the third test does need to be changed in order to be testing the right thing. Changing the first line to `decorator = Other.new.decorate` should do the trick. That way the object _is a_ `Decorator`. Right now that test is only passing because the object we are passing is not a `Decorator`.

2. I think we should add a test case for a non-decorated object:

```ruby
it 'is false for a non-decorator that responds to object' do
  decorator = double(object: Product.new)
  expect(Product === decorator).to be_falsey
end
```

This last test will test the actual change you made (e.g. the object _must be_ a `Decorator`). What do you think?

jrafanie:
> Looks good! Do you think this pull request solves the issue that #774 is trying to solve? I believe it would, but wanted to get a second opinion..

yes, @codebycliff i think so.  It's much safer to delegate to our method for our type of objects.  Otherwise, arity our other interface differences on other objects could give us headaches.

jrafanie:
I'll try to fix the tests shortly

jrafanie:
@codebycliff I totally agree with your suggestions on the test changes.  I pushed the changes and amended the commit.  Let me know what you think.

PR #723, opened 2016-02-14
Implement #deep_dup

textgoeshere:
`Decorator#deep_dup` `dup`-s the decorator and `deep_dup`s its `object`.

Any interest?

One drawback of this approach is that I've had to add a (protected) setter for `object` onto `Decorator`. Perhaps there's a neater way of getting the same result?

### DIFF
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -232,6 +232,13 @@ module Draper
       object.attributes.select {|attribute, _| respond_to?(attribute) }
     end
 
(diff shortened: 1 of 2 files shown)


codebycliff:
Is there a use case you had in mind for this? or just an idea? If it's something you actually use or need, I would be happy to merge it.

codebycliff:
I'm going to close this for now. I'd be happy to re-visit it if it's something you'd like to discuss in the future.

Issue #704, opened 2015-11-03
Test classes loaded unecessarily


drock:
This is related to #690 but not the same.

If the Rails.env.test? is true then Draper::Railtie is automatically bringing in draper/test_case.  However, just because the environment is "test" doesn't necessarily mean that we are running tests and need the test cases classes loaded.  This is particularly true with systems such as CircleCi which by default will export RAILS_ENV="test" and then run various rake tasks.

This is particularly a problem with Ruby 2.2 and Rails 3.2.  The inclusion of draper/test_case then brings in active_support/test_case which in turn loads test/unit.  The unbundled test-unit gem installs an at_exit hook to auto run the tests.  Now, we have any rake task trying to auto run a test unit suite and it bombs.

This example is really more of a side effect of the higher level problem which is that the test library is being loaded even if we aren't necessarily running tests.

codebycliff:
I'm running the latest Draper with Rails 5 in CircleCI with no problem. Draper no longer support Rails 3 / test-unit, which appears to be more of the issue. I also think it's a fair assumption that it's okay to load a libraries test classes when the Rails environment is test.

Issue #679, opened 2015-06-19
Suddenly register_spec_type error


hendricius:
Hi everyone,

thanks for the great gem. Suddenly out of nothing I am receiving a `register_spec_type` error. No idea, for my colleagues it works, no longer on my machine.

I have the draper `2.1.0` version with ruby `2.2.2` and rails `4.2.2`

Any idea what could cause this?

Here is the full stacktrace.

```
➜  tripl_app git:(staging) bin/rake --trace
** Invoke default (first_time)
** Invoke test (first_time)
** Execute test
** Invoke test:run (first_time)
** Execute test:run
rake aborted!
NoMethodError: undefined method `register_spec_type' for Draper::TestCase:Class
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/draper-2.1.0/lib/draper/test/minitest_integration.rb:2:in `<class:TestCase>'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/draper-2.1.0/lib/draper/test/minitest_integration.rb:1:in `<top (required)>'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.2/lib/active_support/dependencies.rb:274:in `require'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.2/lib/active_support/dependencies.rb:274:in `block in require'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.2/lib/active_support/dependencies.rb:240:in `load_dependency'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.2/lib/active_support/dependencies.rb:274:in `require'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/draper-2.1.0/lib/draper/railtie.rb:56:in `block (2 levels) in <class:Railtie>'
/Users/hendricius/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/activesupport-4.2.2/lib/active_support/lazy_load_hooks.rb:38:in `instance_eval'
[...]

PR #726, opened 2016-02-24
Remove loading test code when loading draper.

bcavileer:
@drock https://github.com/drapergem/draper/issues/704

Removed 
      if Rails.env.test?
        require 'draper/test_case'
        require 'draper/test/rspec_integration' if defined?(RSpec) and RSpec.respond_to?(:configure)
      end

### DIFF
--- a/lib/draper/railtie.rb
+++ b/lib/draper/railtie.rb
@@ -16,11 +16,6 @@ module Draper
 
     config.after_initialize do |app|
       app.config.paths.add 'app/decorators', eager_load: true
-
-      if Rails.env.test?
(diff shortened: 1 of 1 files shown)


codebycliff:
Looks like this is breaking the build, particularly while running `rake` in the dummy app. If you would like to try and fix the build, I would happily accept another pull request for this. The error that's breaking the build looks something like:

> minitest_integration.rb:2:in `<class:TestCase>': undefined method `register_spec_type' for Draper::TestCase:Class (NoMethodError)

which could be related to https://github.com/drapergem/draper/issues/679.

PR #728, opened 2016-02-28
add simplified chinese translation for README;

zamia:
README is great for learn decorator/drapper, so I translate it to Simplified Chinese(简体中文) for chinese users;

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -1,0 +1,487 @@
+# Drapper: View Models for Rails
+
+Drapper 使用『面向对象』的方式，给 Rails 添加了独立的视图层。
+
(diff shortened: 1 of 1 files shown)


seanlinsley:
Thank you so much for translating the readme. One problem is it will be hard to keep the English and Chinese versions in sync. I'm not sure how we could manage that.

codebycliff:
While I hate to close a pull request that adds translation, I agree with Sean's concerns about keeping it up to date. I'm not sure how we would could manage this well either. I'm going to close this for the time being; however, if you have suggestions on how to keep it up to date or feel strongly that it should be added, please let me know and I'd be happy to revisit this conversation.

varyonic:
Wiki maybe?

codebycliff:
Yes, absolutely. That was my first thought, but was unsure because it still suffers from the same problem of keeping up to date.

@zamia I have added your translation to the wiki [here](https://github.com/drapergem/draper/wiki/README-(Simplified-Chinese-Translation)).

PR #738, opened 2016-04-19
Fixed deprecation warning: before_filter will be removed in Rails 5.1

micapam:
Gets rid of deprecation warning for before_filter while providing a fallback for old Rails versions.

### DIFF
--- a/lib/draper.rb
+++ b/lib/draper.rb
@@ -32,8 +32,8 @@ module Draper
       include Draper::ViewContext
       extend  Draper::HelperSupport
       extend  Draper::DecoratesAssigned
-
-      before_filter :activate_draper
+      
+      public_send(respond_to?(:before_action) ? :before_action : :before_filter, :activate_draper)
     end
   end
 


micapam:
Hmm. Looking at the errors in the builds, they don't seem to relate to my change. They seem to be a problem with Travis: https://github.com/travis-ci/travis-ci/issues/5239

olivierlacan:
@micapam Can you add the following to the `.travis.yml` file?

```ruby
before_install:
  - gem install bundler
```

This should solve the issue for now. I've tried to ping the Travis team on Twitter to get them to update the default Bundler version they're using.

clemens:
Since Draper 3.0 requires Rails >= 5.0, I think the fallback for `before_filter` isn't necessary.

wisetara:
Seems like you could close this bad boy, no?

codebycliff:
Yes. Now that Draper only supports Rails 5, this issue can be closed.

PR #739, opened 2016-04-19
Also initialize ViewContext for rails runner

graaff:
Initialize Draper's ViewContext when using the runner command in a
similar way to using the console command. Without this code run by the
rails runner command (e.g. when processing commands from a message
queue) will not have a properly initialized ViewContext, and
consequently the helpers will not be initialized.

### DIFF
--- a/lib/draper/railtie.rb
+++ b/lib/draper/railtie.rb
@@ -63,6 +63,12 @@ module Draper
       Draper::ViewContext.build
     end
 
+    runner do
+      require 'action_controller/test_case'
+      ApplicationController.new.view_context
+      Draper::ViewContext.build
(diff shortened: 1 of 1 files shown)


graaff:
Some backend code in our application runs out of a "rails runner" command, triggered by a message queue. After upgrading to draper 2.x this code was failing due to the helpers not being available properly, as evidenced by the " undefined method `host' for nil:NilClass" error in the url_helpers. My solution in this pull request is to initialize the "runner" command in the same way as the "console" command.

codebycliff:
@graaff Thanks for the contribution! This is in master now and will be in our upcoming 3.0.0 release.

PR #740, opened 2016-04-20
Add projections for rails-vim

eloyesp:
Add projections.json to make [vim-rails](https://github.com/tpope/vim-rails) know about decorators.

### DIFF
--- a/lib/rails/projections.json
+++ b/lib/rails/projections.json
@@ -1,0 +1,12 @@
+{
+  "app/decorators/*_decorator.rb": {
+    "command": "decorator",
+    "related": "app/models/{}.rb",
+    "affinity": "model",
+    "template": [
+      "class {camelcase|capitalize|colons}Decorator < Draper::Decorator",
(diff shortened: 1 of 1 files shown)


seanlinsley:
What does this file do, exactly? What are projections?

eloyesp:
Projections are a way to make vim-rails (or any IDE capable of reading the file) about the project making possible to navigate between related files, running the right tests and knowing the existing words.

eloyesp:
Any news about this?

codebycliff:
@eloyesp Thanks for the contribution and sorry it took so long to get back to you. I don't feel like editor specific code belongs in the actual library. I was willing to look past my opinion if any other major gems are doing this; however, it doesn't seem to be a standard practice. What I did do though is add a wiki page with your example [here](https://github.com/drapergem/draper/wiki/Projections-for-rails.vim). This wiki page also includes a link to a Thoughtbot article explaining various ways to set this up.

PR #737, opened 2016-04-06
Rails 5

Mehonoshin:
Lets collect all rails 5 related fixes here
Any help is appreciated :heart:

PR #741, opened 2016-05-27
Rails5 compatibility

kbrock:
Here are the changes we use locally.

They may not solve every draper issue, but they should work across all versions and I was hoping we could get this first batch in.

Thanks,

PS. this builds on some entries from #737 -- I tried to include attribution where posisble

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ services:
 rvm:
   - 2.1.5
   - 2.2.1
+  - 2.3.0
 
 env:
   - "RAILS_VERSION=4.0"
   - "RAILS_VERSION=4.1"
   - "RAILS_VERSION=4.2"
+  - "RAILS_VERSION=5.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,10 @@ env:
   - "RAILS_VERSION=4.1"
   - "RAILS_VERSION=4.2"
   - "RAILS_VERSION=5.0"
+
+matrix:
+   exclude:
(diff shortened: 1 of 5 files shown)


djberg96:
@kbrock Not sure I understand the build failure. What bundler just having a bad day, or does the travis.yml file need tweaking?

DexterTheDragon:
Travis has a really old version of bundler installed by default. Adding the following to your travis.yml will probably fix it.

```yaml
before_install:
  - gem update bundler
```

djberg96:
@DexterTheDragon Now we're seeing all kinds of (seemingly) unrelated errors.

DexterTheDragon:
Looks like various errors in the test suite between all the different versions. I would bet that even the master branch doesn't pass now. If you look at the output from the last time the master branch was tested ( https://travis-ci.org/drapergem/draper/jobs/56058601 ), it was using capybara 2.4.4. Now it's using capybara 2.7.1 which probably needs a change in configuration.

djberg96:
@DexterTheDragon It looks like there hasn't been an update on master in over a year. Doesn't bode well.

seanlinsley:
@djberg96 I recently gained commit access to Draper, to get it working with Rails 5 🙌

djberg96:
@seanlinsley Any updates?

seanlinsley:
In #752 I've put together all the changes necessary to get Draper working with Rails 5. It would be appreciated if everyone could test it out in their apps and report any issues you find :octocat:

kbrock:
thanks @seanlinsley

PR #743, opened 2016-06-01
Prevent automatically delegated methods from piling up on Decorator class.

sylvain-8422:
This should be a simple and safe way to prevent the issue described in _Pull Request_ #708 .

Decorator classes are used to wrap all sorts of objects of different classes, and right now, delegate methods are added permanently to the original Decorator class, making the code believe that the decorated object can `respond_to?` a lot of things, but when time comes to call the methods, they are not really there, and everything comes crashing down.

Adding delegates to the `singleton_class` should fix the issue, without undesirable side effects.

### DIFF
--- a/lib/draper/automatic_delegation.rb
+++ b/lib/draper/automatic_delegation.rb
@@ -6,7 +6,7 @@ module Draper
     def method_missing(method, *args, &block)
       return super unless delegatable?(method)
 
-      self.class.delegate method
+      self.singleton_class.delegate method
       send(method, *args, &block)
     end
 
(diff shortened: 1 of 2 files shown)


seanlinsley:
@sylvain-8422 thanks for the PR 😄 can you add a test for this?

sylvain-8422:
Yes! I will write one right now!

sylvain-8422:
I added a test. The test would fail on current `master` branch.

I hope this will help fix the issue! :-)

seanlinsley:
👍 thanks

PR #747, opened 2016-06-14
Extend the "Proxying Class Methods" section

rvierich:
Quick addition to the readme to explain how to access class methods defined on the decorator when an association with the model has been set up correctly.


seanlinsley:
Thanks 😄

PR #749, opened 2016-06-21
Update bundler to work with rubygems 2.5+

jrafanie:
Fixes undefined method `spec' for nil:NilClass

See https://github.com/bundler/bundler/issues/3558 for more details

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ rvm:
   - 2.1.5
   - 2.2.1
 
+before_install:
+ - gem update bundler
+
 env:
   - "RAILS_VERSION=4.0"
   - "RAILS_VERSION=4.1"


seanlinsley:
Isn't the correct solution to upgrade Bundler?

jrafanie:
@seanlinsley there are several problems, I'll mark this as WIP to fix all that needs to be done.

jrafanie:
ok @seanlinsley this fixes the undefined method `spec' error.  The others are more problematic.

jrafanie:
I see you have a rails-5 branch that looks like you're trying to get things back in shape.  Should I open a PR against that branch?

Note, I had to do some other hacks to try to fix the `undefined method join for nilclass` due to a nil Rails.root.

```diff
diff --git a/spec/spec_helper.rb b/spec/spec_helper.rb
index df3aad9..a9802c5 100755
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -39,3 +42,8 @@ end
 def protect_module(mod)
   before { stub_const mod.name, mod.dup }
 end
+
+# Create a dummy rails application
+require 'rails/all'
+stub_app = Class.new(Rails::Application)
+stub_app.config.secret_key_base = "some_secret_key_base"
```

That felt really dirty but it gets it down to 3 failures.

PR #750, opened 2016-06-28
Fix before_filter deprecation warning

natebird:
I'm not sure why it keeps pulling in the other two commits. I forked and worked off of the rails-5 branch.


seanlinsley:
Hi, thanks for the PR! In #752 I've put together all the changes necessary to get Draper working with Rails 5. It would be appreciated if you could test it out in your app and report any issues you find :octocat:

PR #751, opened 2016-07-01
replace before_filter with before_action

wintersolutions:
rails 5 deprecation warning was:
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead. (called from block in setup_action_controller at [..]/gems/draper-4b4f392826c4/lib/draper.rb:36)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.gem
-*.rvmrc
+.rvmrc
+.ruby-version
+.ruby-gemset
 .bundle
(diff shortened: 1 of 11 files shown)


seanlinsley:
Hi, thanks for the PR! In #752 I've put together all the changes necessary to get Draper working with Rails 5. It would be appreciated if you could test it out in your app and report any issues you find :octocat:

PR #752, opened 2016-07-04
Rails 5 compatibility

seanlinsley:
Note that for simplicity's sake, this PR drops support for Ruby < 2 and Rails < 5.

It also makes Active Model Serializers a runtime dependency. I'll try and undo that if I can.

### DIFF
--- a/lib/draper/test_case.rb
+++ b/lib/draper/test_case.rb
@@ -29,10 +29,10 @@ module Draper
   end
 end
 
-class ActionController::TestCase
-  include Draper::TestCase::ViewContextTeardown
+if defined? ActionController::TestCase
+  ActionController::TestCase.include Draper::TestCase::ViewContextTeardown
 end
 
-class ActionMailer::TestCase
-  include Draper::TestCase::ViewContextTeardown
+if defined? ActionMailer::TestCase
+  ActionMailer::TestCase.include Draper::TestCase::ViewContextTeardown
 end
--- a/lib/draper/test_case.rb
+++ b/lib/draper/test_case.rb
@@ -29,14 +29,10 @@ module Draper
   end
 end
 
-if defined?(ActionController::TestCase)
-  class ActionController::TestCase
-    include Draper::TestCase::ViewContextTeardown
-  end
+class ActionController::TestCase
+  include Draper::TestCase::ViewContextTeardown
 end
 
-if defined?(ActionMailer::TestCase)
-  class ActionMailer::TestCase
-    include Draper::TestCase::ViewContextTeardown
-  end
+class ActionMailer::TestCase
+  include Draper::TestCase::ViewContextTeardown
 end
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.gem
-*.rvmrc
+.rvmrc
+.ruby-version
+.ruby-gemset
 .bundle
 Gemfile.lock
 pkg/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,5 @@ services:
   - mongodb
 
 rvm:
-  - 2.1.5
-  - 2.2.1
-
-env:
-  - "RAILS_VERSION=4.0"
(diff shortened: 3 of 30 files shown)


shaneog:
Thanks @seanlinsley!
Sadly I can't test this yet, since it depends on the latest version of AMS, and I'm pinned to 0.9.5 since I have a *lot* of code to be changed before I can move to 0.10

seanlinsley:
Okay. I'm currently trying to pin down where exactly Draper interfaces with AMS, as well as activemodel-serializers-xml, to determine if we can just list them in the README or if they actually need to be runtime dependencies.

seanlinsley:
@shaneog I reverted that change, so now AMS isn't required at runtime. The only question is whether [the initializer](https://github.com/drapergem/draper/pull/752/files#diff-5c10d205a9e045ffc4abaf5a33515c26L46) is necessary for 0.9.x. Could you try this branch out on your app?

lib/draper/test_case.rb
```diff
@@ -29,14 +29,10 @@ module Behavior
...
 
-if defined?(ActionController::TestCase)
-  class ActionController::TestCase
-    include Draper::TestCase::ViewContextTeardown
-  end
+class ActionController::TestCase
+  include Draper::TestCase::ViewContextTeardown
 end
 
-if defined?(ActionMailer::TestCase)
```
alexceder: May I ask why you are making this change? It breaks testing apps without ActionMailer included, right?

```
NameError: uninitialized constant ActionMailer
  <hidden>/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/draper-de70008a5e68/lib/draper/test_case.rb:36:in `<top (required)>'
  <hidden>/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/draper-de70008a5e68/lib/draper/railtie.rb:21:in `block in <class:Railtie>'
  etc..
```

shaneog:
@seanlinsley Apologies for the delay - this looks like it works ok for me, although I'm still waiting on another gem before I can do in-depth testing.

lib/draper/test_case.rb
seanlinsley: My mistake, I didn't realize that this file was something that other people's test suites relied upon. I'll undo this change.

bdmac:
@seanlinsley trying to use draper's rails-5 branch with our app and we get this when trying to run `rails c`: 
```
User./Users/bdmac/.gem/ruby/2.3.0/bundler/gems/draper-3c165152041f/lib/draper/view_context.rb:8:in `view_context': super: no superclass method `view_context' for #<ApplicationController:0x007fa8220734c0> (NoMethodError)
```
Note that our ApplicationController extends from ActionController::API instead of ActionController::Base and API does not appear to have view_context available. If I switch to extending ActionController::Base I can load the console but of course that's not great since rails 5 brings rails-api in directly for API-only use cases. I thought I saw some other older PRs that tried to address this view_context problem when using the old rails-api gem w/rails 4x but for whatever reason it does not work with rails 5.

seanlinsley:
@bdmac that doesn't seem related to my changes, and I'm not sure how to fix it without more of the stack trace. Could you open a ticket for that issue with more details?

bdmac:
Sure can although repro should be pretty easy, just switch your application controller to inherit from API instead of Base and try to open rails console. Will get stack trace in a bit.

kbrock:
If you are going to drop rails 4.x, you may want to change the major version, so people still running 4.x can easily peg their gemfile to a draper version

seanlinsley:
@kbrock yep, the plan is to bump to 3.0.0

seanlinsley:
I just got access to Rubygems so I'm going to merge this and publish a prerelease version.

seanlinsley:
3.0.0.pre1 has been released

kbrock:
yay!

natebird:
I'll check out the prerelease this week and report back if I see any issues. Thanks!

PR #762, opened 2016-10-17
Use non-monkey patch describe in generators.

dalpo:
The default Rails spec generators should not use the monkey patched version of
describe. (https://github.com/rspec/rspec-rails/pull/1052)


seanlinsley:
👍 thanks

PR #763, opened 2016-10-31
Bugfix for missing ActionController::TestRequest

McRip:
Fixed a bug where helpers were used inside a decorator and this decorator was used outside of controller context.


seanlinsley:
This needs to be rebased. Also, could you add a test?

McRip:
I am sorry, but i see no way to test this, because ActionController::TestRequest is always loaded inside RSpec.

alessandro1997:
Are there any updates on this?

McRip:
Unfortunately not. @seanlinsley, can this be merged?

PR #765, opened 2016-11-25
minor fix to error raising on collection decoration.

almaron:
The previous version raised an error with eager loaded classes in case one had say a model class with the same name as the decorator module and trying to decorate a collection. 
E.g. having an `Article` model, an `Article::AuthorCredential` model and an `Article::AuthorCredentialDecorator` would raise a NameError when trying to do `Article::AuthorCredential.all.decorate`.
This fix eliminates the problem

### DIFF
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -231,7 +231,7 @@ module Draper
       name = collection_decorator_name
       name.constantize
     rescue NameError => error
-      raise if name && !error.missing_name?(name)
+      raise if name && !name.include?(error.missing_name)
       Draper::CollectionDecorator
     end
 


seanlinsley:
Thanks for the fix! Could you add a test for this?

codebycliff:
I believe this solves the issue https://github.com/drapergem/draper/issues/773. So the issue is as follows:

When calling `Article::AuthorCredential.all.decorate`, it calls `collection_decorator_class` to try and infer the collection decorator. In this case, it tries to constantize `Article::AuthorCredentialsDecorator` which doesn't exist. The rescue block is designed to re-raise the `NameError` if the `NameError#missing_name` does not match the value we were trying to constantize. However, for some reason the value of `missing_name` in this case is `::AuthorCredentialsDecorator` (without the prefixed namespace). So since the `missing_name` didn't match what we were trying to constantize the `NameError` is re-raised.

The fix provided was to see if the value we were trying to constantize _contained_ the `NameError#missing_name`. While this does indeed fix the issue, I wonder about this leading to false positives. Even more, I question whether that matters at all and if we shoud never re-raise and always default to `CollectionDecorator`. This could still have the side affect of silencing `NameError`s that we didn't cause. But I'm not sure there is a code path to support that scenario and, if there was, what the consequences would be.

Any thoughts? @seanlinsley @jcasimir

codebycliff:
So a little more background on this. The issue only happens when:
```ruby
config.eager_load = true
config.cache_classes = false
config.enable_dependency_loading = false // Turning this on makes the problem go away
```

We could reduce the chances of false positives if we try to `demodulize` first:

```ruby
  rescue NameError => error
      raise if name && name.demodulize != error.missing_name.demodulize
      Draper::CollectionDecorator
 end
```

But false positives could still leak through if you had: `Article::AuthorCredential` and `Research::Article::AuthorCredential`. 

Right now, I think I'm leaning toward always defaulting to `CollectionDecorator` and not re-raising at all. All the code paths that could throw a `NameError` that we wouldn't want to silence would still get hit elsewhere in the decoration process (e.g. when decorating the individual item). Does anyone else have an opinion on what to do here? @syguer?

syguer:
I agree with you :D
`NameError` might be raised anywhere, but not here.
When collection decorator is not found, should return `Draper::CollectionDecorator` here.

codebycliff:
Okay. New pull request has been opened with tests: https://github.com/drapergem/draper/pull/795. Closing this one now.

PR #754, opened 2016-07-13
Support Rails::API controllers

seanlinsley:
#753

Issue #753, opened 2016-07-11
Draper rails-5 + ActionController::API


bdmac:
Unable to load rails console using Rails 5 + the rails-5 branch of draper if my ApplicationController inherits from `ActionController::API` instead of `ActionController::Base`. Full stack trace below.

Steps to repro:
1. Take any Rails 5 app using draper.
2. Change ApplicationController parent class to be `ActionController::API` instead of `ActionController::Base`.
3. Run `rails console`.

```
/Users/bdmac/.gem/ruby/2.3.0/bundler/gems/draper-3c165152041f/lib/draper/view_context.rb:8:in `view_context': super: no superclass method `view_context' for #<ApplicationController:0x007ff0e76167c0> (NoMethodError)
Did you mean?  view_runtime
    from /Users/bdmac/.gem/ruby/2.3.0/bundler/gems/draper-3c165152041f/lib/draper/railtie.rb:62:in `block in <class:Railtie>'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/railtie.rb:226:in `block in run_console_blocks'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/railtie.rb:247:in `each'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/railtie.rb:247:in `each_registered_block'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/railtie.rb:226:in `run_console_blocks'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/application.rb:463:in `block in run_console_blocks'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/engine/railties.rb:13:in `each'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/engine/railties.rb:13:in `each'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/application.rb:463:in `run_console_blocks'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/engine.rb:442:in `load_console'
    from /Users/bdmac/.gem/ruby/2.3.0/gems/railties-5.0.0/lib/rails/commands/console.rb:34:in `initialize'
[...]

PR #766, opened 2016-11-29
Support Rails 5 API-only apps

jpettettphaxio:
This is a partial fix for #753 and #754.  Including `ActionView::Rendering` in `ApplicationController` fixes the `view_context` `NoMethodError`.  However, doing so currently breaks any actions which use `render :json` (may be a Rails bug, see rails/rails#27211).  This PR allows an alternate controller to be used by Draper so that ApplicationController is not affected.

I'm aware that this is a hackish solution, and I'm hoping that the Rails issue will be fixed and make this PR mostly irrelevant.  But I'm creating this here because it does work for us in the mean time and it may lead to a better understanding of the problem and a cleaner solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,18 @@ And run `bundle install` within your app's directory.
 If you're upgrading from a 0.x release, the major changes are outlined [in the
 wiki](https://github.com/drapergem/draper/wiki/Upgrading-to-1.0).
 
+### Rails 5 API-only compatibility
+
+Draper expects your `ApplicationController` to include `ActionView::Rendering`.  However, Rails 5 API-only applications (created with `rails new --api`) don't by default.  However, including `ActionView::Rendering` in ApplicatonController also breaks `render :json`.  Therefore, you use an alternate controller by adding the following to your `application.rb`:
+
+```ruby
+draper__view_context = ::Draper::ViewContext
+def draper__view_context.controller
+  Class.new(ApplicationController) { include ActionView::Rendering }.new
+end
+
+```
+
 ## Writing Decorators
 
 Decorators inherit from `Draper::Decorator`, live in your `app/decorators`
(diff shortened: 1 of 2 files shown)


codebycliff:
@jpettettphaxio Thanks for the contribution. I'd like to get support for API only apps in the next release. I'm waiting to get https://github.com/drapergem/draper/pull/776 in first, as that will make customizing the default controller easier. 

As far as breaking `render :json`, how do you think we should proceed? I checked the rails issue and there doesn't seem to be much attention there. So I doubt we can expect a patch anytime soon. 

Assuming the pull request above is merged, we could update the README with something like:

```ruby
Draper::ViewContext.default_controller = Class.new(ActionController::API) do
  alias_method :_orig_render_to_body, :render_to_body
  include ActionView::Rendering
  alias_method :render_to_body, :_orig_render_to_body
end
```

Would that make sense?

codebycliff:
As soon as https://github.com/drapergem/draper/pull/788 is merged the documentation would read:

> ...Therefore, you use an alternate controller by adding the following to `config/initializers/draper.rb`:
> ```ruby
> Draper.configure do |config|
>   config.default_controller = Class.new(ApplicationController) do
>     alias _orig_render_to_body render_to_body
>     include ActionView::Rendering
>     alias render_to_body _orig_render_to_body
>   end
> end
> ```

I have verified the above works in combination with https://github.com/drapergem/draper/pull/788 for Rails API only applications. It solves issue https://github.com/drapergem/draper/issues/753 as well. This will be the solution for now. I may provide extra support to make it easier in the future (maybe an `api_only` configuration option that handles this setup).

Let me know if you want to update the pull request or if I should just just branch and take it over.

codebycliff:
Instead of just documentation, I decided to build this patch in so that Rails 5 API-only apps work out of the box. You can find the pull request [here](https://github.com/drapergem/draper/pull/793). Thanks for you help getting to the bottom of the issue. I'm going to close this pull request now.

Issue #725, opened 2016-02-22
`source` method name conflicts with AR relation


troex:
I have a model:

``` ruby
module Shop
  class Click < ActiveRecord::Base
    belongs_to :source, polymorphic: true
  end
end
```

When model is undecorated calling `click.source` returns AR relation, but if model is decorated it returns self when called. example `click.decorate.source == click` returns `true`

I understand this is internal name conflict, but I couldn't find any info on "reserved" method names. Any luck getting AR relation without changing relation name?

troex:
References #501, solution is provided but old alias methods still present

troex:
```ruby
module Draper
  class Decorator
    remove_method :source
  end
end
```

worked for me as a temporary solution

wilkenm:
Just spent a few hours dealing with this issue, +1 for getting rid of the 'source' alias now that it's not used internally.

codebycliff:
This is in master now.

PR #768, opened 2016-12-06
Rename UninferrableSourceError to UninferrableObjectError

ertrzyiks:
Since source has been renamed to object, it is better for consistency to have UninferrableObjectError class name with the error message mentioning 'object' property instead of 'source'.

### DIFF
--- a/lib/draper.rb
+++ b/lib/draper.rb
@@ -55,9 +55,11 @@ module Draper
     end
   end
 
-  class UninferrableSourceError < NameError
+  class UninferrableObjectError < NameError
     def initialize(klass)
-      super("Could not infer a source for #{klass}.")
+      super("Could not infer an object for #{klass}.")
     end
   end
+
+  UninferrableSourceError = UninferrableObjectError # TODO: deprecate this
 end
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -72,7 +72,7 @@ module Draper
     # Checks whether this decorator class has a corresponding {object_class}.
     def self.object_class?
       object_class
-    rescue Draper::UninferrableSourceError
+    rescue Draper::UninferrableObjectError
       false
     end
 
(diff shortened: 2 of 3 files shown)


seanlinsley:
Wouldn't this also make sense?
```ruby
class UninferrableSourceError < NameError
  # ...
end

UninferrableSourceError = UninferrableObjectError # TODO: deprecate this
```

ertrzyiks:
@seanlinsley did you mean `class UninferrableObjectError < NameError`? makes sense to me
If you want I can do this in the scope of this PR as well

seanlinsley:
Yeah that's what I meant 😄 It'd also be good to update any internal uses to `UninferrableObjectError`.

ertrzyiks:
@seanlinsley updated according to your comment

For the record: CI check for this PR was red once: https://travis-ci.org/drapergem/draper/builds/192448535 seems to be some random issue, I've tried to reproduce it locally but `bundle exec rake` fails on my machine with an error about missing Mongo connection (totally valid point though).

codebycliff:
@ertrzyiks Thanks for the contribution. Inconsistency is the worst! I'm going to go ahead and merge this as is. I'm thinking about removing the backward compatibility (`UninferrableSourceError = UninferrableObjectError`) as well since this will be going in a major release. Do you see any major issue in doing so?

varyonic:
Would something like this work?
```ruby
class UninferrableSourceError < NameError
  def intialize
    ActiveSupport::Deprecation.warn 'Use UninferrableObjectError ', caller
    super
  end
end
```

codebycliff:
Yeah, I wasn't sure how we wanted to handle these name changes. We could either:

1. Deprecate them with active support and remove them in 3.1.0. I would think this would be the best practice, however some of these name changes are actual conflicts (https://github.com/drapergem/draper/issues/725), so it might benefit from being removed entirely.

2. Just remove them and list them as a breaking change in the changelog. Not the best practice, but neither is upgrading to a major release of gem without reading the changelog.

What do you think?

varyonic:
Ah, remove it then.

PR #774, opened 2017-01-16
Fix issue with decoratable class equality operator

abrom:
If you try compare a decorated class with an object that responds to object *but requires arguments*, the overridden class equality operator will cause an exception to be raised. This change requires that the `other` object method arity be 0 (no args) or -1 (any number of args)

The specific place where I've found this to be an issue is when using the rspec `(expect|allow)_any_instance_of` method and passing it an ActiveRecord model in combination with the redis gem (where the redis client class responds to object but requires an argument). Rspec tries to set up its mock proxies where it then compares the AR class to the redis instance. *doh*

I've included a spec that validates that the operator now just returns false for that use case

### DIFF
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -86,7 +86,11 @@ module Draper
       #
       # @return [Boolean]
       def ===(other)
-        super || (other.respond_to?(:object) && super(other.object))
+        super || (
+          other.respond_to?(:object) &&
+            [-1, 0].include?(other.method(:object).arity) &&
+            super(other.object)
+        )
       end
 
     end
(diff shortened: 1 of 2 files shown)


abrom:
One thought would be to not include the arity -1 option and only compare when the object method is arity 0. Although doing so will break the specs as the mocks are -1.

codebycliff:
@abrom Thanks for the contribution! I think this issue is actually solved by https://github.com/drapergem/draper/pull/720. That pull request changes the `===` to ensure the other object `is_a? Decorator`. This would eliminate the scenario where the other object doesn't have an `object` method with arity 0.

If I'm misunderstanding the issue or incorrect about the other pull request solving the problem, let me know and I'd be happy to re-open the pull request.

Issue #573, opened 2013-09-10
Railtie requires a controller named ApplicationController


wgrrrr:
The railtie for Draper appears to require a controller named ApplicationController to exist. This is a big problem for people (like myself) who don't use the standard naming convention for the base controller.

I believe this should be configurable to allow non-standard naming conventions.

https://github.com/drapergem/draper/blob/master/lib/draper/railtie.rb#L55

Thoughts?

```
/Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/draper-1.2.1/lib/draper/railtie.rb:55:in `block in <class:Railtie>': uninitialized constant Draper::Railtie::ApplicationController (NameError)
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/railtie.rb:188:in `call'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/railtie.rb:188:in `block in run_console_blocks'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/railtie.rb:188:in `each'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/railtie.rb:188:in `run_console_blocks'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/application.rb:264:in `block in run_console_blocks'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/engine/railties.rb:17:in `each'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/engine/railties.rb:17:in `each'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/application.rb:264:in `run_console_blocks'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/engine.rb:431:in `load_console'
    from /Users/me/.rvm/gems/ruby-2.0.0-p247@gemset/gems/railties-4.0.0/lib/rails/commands/console.rb:51:in `initialize'
[...]

Issue #249, opened 2012-08-08
uninitialized constant Draper::Railtie::ApplicationController


krainboltgreene:
```
$ rails console
Connecting to database specified by database.yml
1.9.3-p194@manacurve-api/gems/rake-0.9.2.2/lib/rake/ext/module.rb:36:in `const_missing': uninitialized constant Draper::Railtie::ApplicationController (NameError)
  from 1.9.3-p194@manacurve-api/gems/draper-0.15.0/lib/draper/railtie.rb:39:in `block in <class:Railtie>'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/railtie.rb:179:in `call'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/railtie.rb:179:in `block in load_console'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/railtie.rb:179:in `each'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/railtie.rb:179:in `load_console'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/engine.rb:429:in `block in load_console'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/application/railties.rb:8:in `each'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/application/railties.rb:8:in `all'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/engine.rb:429:in `load_console'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/application.rb:153:in `load_console'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/commands/console.rb:27:in `start'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/commands/console.rb:8:in `start'
  from 1.9.3-p194@manacurve-api/gems/railties-3.2.7/lib/rails/commands.rb:41:in `<top (required)>'
  from script/rails:6:in `require'
  from script/rails:6:in `<main>'
```

I want to say I've reported this before, but I can't find it.

Basically Draper doesn't function in the console without an ApplicationController. My API doesn't have an ApplicationController. I only have namespaces (versions) so the class I have is V1::ApplicationController.

I really feel like this was already solved, but I again I can't find anything.

PR #776, opened 2017-02-10
Allow for configuration of the default view context controller 

nick-donald:
In response to issues https://github.com/drapergem/draper/issues/249 and https://github.com/drapergem/draper/issues/573

Allow for configuration of the default view context controller. This is particularly useful for applications without an `ApplicationController` (ie engine-driven apps). A string is used rather than a constant so that the `default_controller` can be set anywhere in the initialization process


codebycliff:
@nick-donald Sorry it took so long to get back to you. Thanks for the contribution. It looks like this has been a pain point for a while now. I was able to test this by adding a new initializer containing `Draper::ViewContext.default_controller = 'BaseController'`. Is that the process you were imagining? Before I merge this, I'd like to see a few more changes:

1. Update the Railtie's runner to use the new `ViewContext.default_controller`
2. Add tests
3. Add documentation to the README.

Would you be able to take a stab at these things?

codebycliff:
@nick-donald I have continued your work in the following pull request:

https://github.com/drapergem/draper/pull/788

Thanks for your help!

nick-donald:
sure 👍

PR #785, opened 2017-03-29
Remove previously deprecated functionality.

codebycliff:
This pull request removes all functionality that was previously deprecated. This code was originally in [Support Rails::API controllers](https://github.com/drapergem/draper/pull/754), but I thought it would be best to pull it out into a separate pull request for better tracking.


syguer:
> I thought it would be best to pull it out into a separate pull request for better tracking.

I agree. Looks good 👍

PR #788, opened 2017-03-30
Add support for custom default controller configuration

codebycliff:
## Description
This pull request adds configuration to draper so that you can customize the default controller. According to the documentation in the README:

> ## Configuration
> Draper works out the box well, but also provides a hook for you to configure its 
> default functionality. For example, Draper assumes you have a base `ApplicationController`.
> If your base controller is named something different (e.g. `BaseController`),
> you can tell Draper to use it by adding the following to an initializer:
>
>```ruby
>Draper.configure do |config|
>  config.default_controller = BaseController
>end
>```

This also lays the ground work for other configuration options in the future. This work was originally inspired from the [#776](https://github.com/drapergem/draper/pull/776) by @nick-donald.

## Todos
- [x] Tests
- [x] Documentation
- [x] Fix state leaking from tests

## Related

Issue https://github.com/drapergem/draper/pull/776
Issue https://github.com/drapergem/draper/issues/573
Pull request https://github.com/drapergem/draper/pull/776

### DIFF
--- a/spec/draper/configuration_spec.rb
+++ b/spec/draper/configuration_spec.rb
@@ -1,0 +1,25 @@
+require 'spec_helper'
+
+module Draper
+  RSpec.describe Configuration do
+    it 'yields Draper on configure' do
+      Draper.configure { |config| expect(config).to be Draper }
+    end
+
+    it 'defaults default_controller to ApplicationController' do
+      expect(Draper.default_controller).to be ApplicationController
+    end
+
+    it 'allows customizing default_controller through configure' do
+      default = Draper.default_controller
+
+      Draper.configure do |config|
+        config.default_controller = CustomController
+      end
+
+      expect(Draper.default_controller).to be CustomController
+
+      Draper.default_controller = default
+    end
+  end
+end
--- a/spec/draper/configuration_spec.rb
+++ b/spec/draper/configuration_spec.rb
@@ -7,10 +7,12 @@ module Draper
     end
 
     it 'defaults default_controller to ApplicationController' do
+      skip
       expect(Draper.default_controller).to be ApplicationController
     end
 
     it 'allows customizing default_controller through configure' do
+      skip
       default = Draper.default_controller
 
       Draper.configure do |config|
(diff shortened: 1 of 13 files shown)


spec/draper/configuration_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+module Draper
+  RSpec.describe Configuration do
+    it 'yields Draper on configure' do
+      Draper.configure { |config| expect(config).to be Draper }
+    end
+
+    it 'defaults default_controller to ApplicationController' do
+      skip
```
codebycliff: These tests all pass every time when I run just this file. However, these last two tests cause the entire suite to fail with an error about leaking a double. Due to it being a state leaking issue, it's not a consistent test that fails.  The two different errors that keep seeing are:
image

and 

image

I'm not sure how to go about fixing those as I'm not super familiar with the test setup on this project. I will continue to look into it, but could use some advice if someone has some.

codebycliff: I was able to fix this issue. The problem was that one of the tests in the `build_strategy_spec.rb` was stubbing a constant that was overriding the new class definition in the `spec_helper.rb`. The specific test that was stubbing was able to be re-written without mocks now that the `ApplicationController` class is defined. I believe the way I re-wrote the test still tests for the same functionality. Somebody correct me if that's not the case. The commit is here: 
https://github.com/drapergem/draper/pull/788/commits/a886d288aea00c09388bc0fd6caac32badda6aaa

codebycliff:
This is ready for review now if anybody wants to take a look (e.g. @varyonic @seanlinsley @syguer). I added the new functionality to the dummy application as well to increase the code coverage.

syguer:
Looks good! 👏

syguer:
But I wonder why draper depends on Controller layer... 🤔 
Of course I know,  we need a context of controller. I hope we can divide its dependency someday 😄

codebycliff:
Yes, I agree. Maybe we can extract that dependency into optional functionality. The basic decorator pattern should be able to work without the need for a view context. But, baby steps for now..

syguer:
👍

PR #789, opened 2017-03-31
Don't do runtime delegation, delegate explicitly

codebycliff:
This was originally pull request https://github.com/drapergem/draper/pull/665, but with the conflicts resolved and the test fixed. From the original pull request description:

> 
> If someone accidentally calls `super` instead of `object` inside a decorator, all hell breaks loose.
> 
> ``` rb
> require "draper"
> require "ostruct"
> 
> class PostDecorator < Draper::Decorator
>   delegate_all
> 
>   def title
>     super and "overriden title"
>   end
> end
> 
> post = OpenStruct.new(title: "title")
> decorated_post = PostDecorator.new(post)
> 
> puts decorated_post.title #=> "overriden title"
> puts decorated_post.title #=> "title"
> ```
> 
> This patch makes both `super` and `object` work. I'm not sure why runtime delegation was added in the first place, but it seems very wrong to add methods on the fly.


syguer:
Is this same issue? https://github.com/drapergem/draper/pull/717

syguer:
I said on [my comment](https://github.com/drapergem/draper/pull/717#issuecomment-290404122), it looks coincidence that `super` calls the method implemented on class of `object`. 
Because parent class (not class of `object`) almost does not have the method(target of `super`). So it calls `method_missing`.

`super` is not for `object`, but for parent class. isn't it?

spec/draper/decorator_spec.rb
```diff
@@ -571,12 +571,16 @@ module Draper
...
+              super and "overriden hello world"
+            end
+          end
+          decorator = decorator_class.new(double(hello_world: "hello world"))
 
-          expect(decorator.methods).not_to include :hello_world
-          decorator.hello_world
-          expect(decorator.methods).to include :hello_world
+          expect(decorator.hello_world).to eq "overriden hello world"
+          expect(decorator.hello_world).to eq "overriden hello world"
```
syguer: It looks same as above.
If code does not delegate anymore, secondary calling is success obviously. So I think this line is unnecessary.

codebycliff:
@syguer Couple of things here:

1. Yes, it the same issue as https://github.com/drapergem/draper/pull/717. I have closed that pull request in favor of this one.

2. Outside of the discussion on `super`, this change makes sense right? Delegating the method explicitly through `send` instead of through `delegate`? Essentially accomplishes the same thing without modifying the class on the fly. I think this change makes sense, but wanted to verify with you.

3. As far as `super` goes, I believe calling `super` from within the decorator should never delegate to object and always attempt to call the parent class. The functionality of `super` is for inheritance, not composition (which is what the decorator pattern is). An example of what I think the expected functionality should be:

```ruby
class Foo
  def title
    'foo title'
  end

  def description
    'foo description'
  end
end

class AppDecorator < Draper::Decorator
  def title
    'app title'
  end
end

class FooDecorator < AppDecorator
  def title
    super
  end

  def description
    super
  end
end

foo = Foo.new
# Should output 'foo title'
puts foo.title

decorated_foo = FooDecorator.new(foo)

# This should output 'app title' because `super` should call parent class.
# If you want the wrapped object's `title`, use `object.title`
puts decorated_foo.title

# This should raise `NoMethodError`, because `description` does not exist
# on the parent class. If you want `Foo#description`, once again you will
# need to use `object.description`
puts decorated_foo.description

```

Do you agree or disagree with this? If you agree, I'll get some tests added to cover this scenario.

syguer:
@codebycliff 
I think this change makes sense too 👍 
And I agree the behaviour of your example. I think it is OK if spec is defined and user can understand behaviour correctly 👍

lib/draper/automatic_delegation.rb
```diff
@@ -6,8 +6,7 @@ module AutomaticDelegation
     def method_missing(method, *args, &block)
       return super unless delegatable?(method)
 
-      self.singleton_class.delegate method
-      send(method, *args, &block)
+      object.send(method, *args, &block)
```
codebycliff: @syguer I'm not sure how to prevent the `super` call from reaching `object` because by the time we hit this `method_missing`, there is no way to tell that the method was called from `super` because it has been resolved to the method name:

image

The runtime exception from the screenshot is due to me explicitly raising in `Post#bar` to get the call stack for demonstration purposes. This doesn't have anything to do with the behavior being discussed.

So as the screenshot illustrates:

1. Calling `super` within `PostDecorator#bar` tries to call `ApplicationDecorator#bar`.
2. This triggers `AutomaticDelegation#method_missing` with the name of the method (`:bar`).
3. At this point, I have no idea that `bar` was called through `super` and it sees that `Post#bar` exists and calls that.


I'm not sure how to stop this behavior. Right now, I'm leaning toward leaving the behavior as is and:

1. Documenting that `super` will try to call a parent class and if the method doesn't exist, it will try to call the method on `object`.
2. Ensuring the tests are covering this functionality
3. Merging this fix as is (since it actually fixes another issue not entirely related to `super`).

What do you think?

syguer: Thank you for trying :)
I understand difficulty of it. ([Kernel#caller](https://docs.ruby-lang.org/en/2.4.0/Kernel.html#method-i-caller) might be able to do. But I think it  is too much supporting)

I agree your suggestion. It looks ok if that behavior covered by black box test. (and has documentation 👍 )

codebycliff: Sounds good. I added documentation to the source and README. I also added some additional tests to cover the scenarios. I'm going to go ahead and merge this then.

Issue #784, opened 2017-03-28
Fix indeterminate test causing builds to fail randomly


codebycliff:
There is one test that fails randomly on Travis:

![image](https://cloud.githubusercontent.com/assets/651946/24421211/0b0429e2-13bb-11e7-9462-2c06eaa74c87.png)

There doesn't seem to be any obvious correlation between the failed builds (e.g. ruby version). You can check the various failed builds [here](https://travis-ci.org/drapergem/draper/builds). Just about every single failed build is due to this one specific test.

varyonic:
Try [rspec --bisect --seed 22655](http://blog.arkency.com/2016/02/how-rspec-helped-me-with-resolving-random-spec-failures/)?

codebycliff:
Tried that to no avail. I also tried the above command (`bin/rails test test/decorators/test_unit/helpers_test.rb`), with the seeds from various failed builds. I cannot seem to reproduce this locally at all. This is really hindering productivity and making hard to actually see at a glance which pull requests actually have builds failing. I'm tempting to comment out / skip the test until I can come back and look into it more. What do you think @varyonic?

varyonic:
@deivid-rodriguez Thoughts?

deivid-rodriguez:
@varyonic `rspec --bisect --seed 22655` is exactly what I would have tried. If the failure is not just order dependent but completely flaky then it requires deeper investigation...

@codebycliff If this is blocking other working you are doing and you currently have no time to investigate, I think your approach is right. Skip it with a big TODO and a firm intention of not letting it stagnate there :)

codebycliff:
Thanks for your guy's input. I will skip the test for the time being and leave the issue open.

codebycliff:
Nevermind. Skipping the test that I though was the culprit, just made another test in that suite fail. I'm not sure what to do from here.

varyonic:
When all else fails `rspec --order defined`

deivid-rodriguez:
@codebycliff Well, now you have more knowledge about the problem... That's good.

@varyonic As I understand it, this is not order related so running the tests in order might not help. But it's worth a try I guess.

There's something I don't understand though. Draper's tests seem to use RSpec, but the failure looks like Minitest... Does the test suite run the generated application test suite somehow?

codebycliff:
Yes, it does from what I understand. The library itself provides test integration with RSpec and Minitest. The dummy application tests both integrations.

codebycliff:
[...]

PR #791, opened 2017-03-31
Fix the indeterminate test issue.

codebycliff:
## Description
This pull request fixes the indeterminate test issue causing builds to fail randomly. I finally got to the bottom of what was causing it. The issue was with the minitest spec tests. A couple of the `describe` blocks didn't have the word "decorator" in them. This prevented minitest from registering the spec type as `Draper::TestCase`. Since the tests weren't registered as a `Draper::TestCase`, it wasn't including `Draper::TestCase::ViewContextTeardown` which defines `teardown` to clear the view context. Since the view context wasn't being cleared after these tests, the next test that called a helper on `ViewContext.current` blew up.

## Testing
Once merged, we should not longer see builds fail randomly. You can test locally by:

1. `$ cd spec/dummy`
2. `SEED=8910 rake` should fail repeatedly on master.
3. Switch to this branch and run the above command and the test suite should pass.

## To-Dos
None

## References
* [GitHub Issue 784](https://github.com/drapergem/draper/issues/784)

### DIFF
--- a/spec/dummy/test/decorators/minitest/view_context_test.rb
+++ b/spec/dummy/test/decorators/minitest/view_context_test.rb
@@ -13,12 +13,12 @@ describe "A decorator test" do
   it_does_not_leak_view_context
 end
 
-describe "A controller test" do
-  subject{ Class.new(ActionController::Base) }
+describe "A controller decorator test" do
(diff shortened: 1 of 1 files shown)


syguer:
wtf...
Great job 👍

Issue #773, opened 2017-01-11
NameError: uninitialized constant for Rails 5


vill:
Hi there.  

```
Ruby 2.3.1p112
Rails 5.0.1
draper 3.0.0.pre1
```

`app/models/client/company/deal.rb`
```
module Client
  class Company
    class Deal < ApplicationRecord
      ....
    end
  end
end
```

`app/decorators/client/company/deal_decorator.rb`
```
module Client
  class Company
    class DealDecorator < Draper::Decorator
      delegate_all
      ....
    end
  end
end
```
For `Client::Company::Deal.all.decorate` or `Client::Company::DealDecorator.decorate_collection(Client::Company::Deal.all)` I am getting the following error in the production environment:
```
NameError: uninitialized constant #<Class:0xb65df564>::DealsDecorator
Did you mean?  Client::Company::DealDecorator
```
This error occurs when the following settings:
```
config.cache_classes = true
config.eager_load    = true
```
Everything works fine if I set the following in the `config/environments/production.rb`:
```
config.enable_dependency_loading = true
```
See the [config.enable_dependency_loading](https://github.com/rails/rails/commit/80b416f5e692ae79f4062f59fed6c7d7648a8af0) for details.

HermannHH:
@vill I'm experiencing this exact same issue. My application works fine in development but fails when deployed to Heroku. I posted a S/O question [http://stackoverflow.com/questions/41583880/error-in-heroku-when-using-rails-5-and-draper-namespace](here). Hopefully someone knows how to fix this. I've tried numerous different solutions.

vill:
@HermannHH, I fix this error so: 
```
@deals = Client::Company::Deal.all.map(&:decorate)
```

spike886:
I had the same problem and solved adding this code to the decorators that have namespaces:
```
def self.collection_decorator_class
    Draper::CollectionDecorator
  end
```

codebycliff:
This was merged in https://github.com/drapergem/draper/pull/795. Closing this now.

PR #795, opened 2017-04-04
Never re-raise NameError and always default to CollectionDecorator

codebycliff:
## Description
This pull request makes it so that `NameError` is never re-raised when calling `collection_decorator_class`. This makes it so that it always defaults to `Draper::CollectionDecorator`. The `NameError` that we would be silencing would have been caused by us to begin with. See the original pull request for full detail:

[minor fix to error raising on collection decoration.](https://github.com/drapergem/draper/pull/765)

## Testing
See the original pull request comments.

## To-Dos
None

## References
* [GitHub Issue 773](https://github.com/drapergem/draper/issues/773)
* [GitHub Pull Request 765](https://github.com/drapergem/draper/pull/765)

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,7 +496,10 @@ end
 
 When your decorator calls `delegate_all`, any method called on the decorator not
 defined in the decorator itself will be delegated to the decorated object. This
-is a very permissive interface.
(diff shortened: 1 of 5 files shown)


syguer:
LGTM 👍

Issue #727, opened 2016-02-26
ApplicationDecorator class


dmitrypol:
A common pattern in Rails is to have ApplicationController, ApplicationMailer, etc.  The class inherits from base and developer generated code inherits from it.  This provides a useful place to implement common functionality.  

We are manually introducing this in our app it but it could be a built-in feature.  Does anyone have ideas / suggestions?  

```
class ApplicationDecorator < Draper::Decorator
  delegate_all
  ...
end
class UserDecorator < ApplicationDecorator
  ...
end
```

anbiniyar:
I did the same thing as well. +1

Mehonoshin:
Nice idea!

francirp:
👍 I do the same thing. I typically implement `currency` and `percent` class methods in my `BaseDecorator`. If you're curious I wrote a post on it here: https://medium.com/@ryan_p_francis/how-decorators-saved-my-rails-app-c5cc18ef7d2e

MrEmelianenko:
Yeap, good idea! We could create rake task `rake draper:install` to create this file automatically.

francirp:
Yeah exactly. Happy to take a first punch at a PR for this if it's something the team wants to pursue. Who makes decisions like this on the Draper team? @seanlinsley what are your thoughts on the subject?

seanlinsley:
👍 sounds like a good idea

codebycliff:
This seems like a good idea. Is anyone still interested in taking a stab at a PR?

chuck-john:
@codebycliff, this is a great idea.  I'll get a branch started.

codebycliff:
This was merged in https://github.com/drapergem/draper/pull/796.

PR #796, opened 2017-04-04
ApplicationDecorator generator

chuck-john:
## Description
This branch introduces an installation generator.  Running `rails generate draper:install` will create an `ApplicationDecorator` class, if none already exists.  The purpose of this change is encourage use of the Rails' `Application` naming convention by creating a parent decorator class from which others extend.  Due to [existing work](https://github.com/drapergem/draper/blob/master/lib/generators/rails/decorator_generator.rb#L17), generated decorators will inherit from `ApplicationDecorator`, if present.  More information on this feature request can be found [here](https://github.com/drapergem/draper/issues/727).

## Testing
1. Run `rails generate draper:install` in the dummy app.
1. Ensure the class `ApplicationDecorator` is created in `app/decorators/application_decorator.rb`.
1. Ensure newly-generated decorators extend from `ApplicationDecorator` when present.

## To-Dos
N/A

## References
* [GitHub Issue 727](https://github.com/drapergem/draper/issues/727)


spec/generators/install/install_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  destination File.expand_path('../tmp', __FILE__)
+
+  before { prepare_destination }
+  after(:all) { FileUtils.rm_rf destination_root }
+
+  describe 'the application decorator' do
+    subject { file('app/decorators/application_decorator.rb') }
+
+    describe 'naming' do
+      before { run_generator %w(YourModels) }
```
codebycliff: I don't think you need the `%w(YourModels)` here since this generator doesn't take any arguments.

lib/generators/draper/templates/application_decorator.rb
```diff
@@ -0,0 +1,11 @@
+class ApplicationDecorator < Draper::Decorator
+  delegate_all
+
+  # Define methods for all decorated objects.
+  # Helpers are accessed through `helpers` (aka `h`). For example:
+  #
+  #   def percent_amount
+  #     h.number_to_percentage object.amount, precision: 2
+  #   end
+
```
codebycliff: Remove empty line.

README.md
```diff
@@ -155,6 +162,8 @@ rails generate decorator Article
 
 ...to create the `ArticleDecorator`.
 
+Generated decorators will inherit from `Draper::Decorator` unless an `ApplicationDecorator` exists.
```
codebycliff: I think it might be more in context to move this up with the rest of the new documentation. Something like:

> To create an `ApplicationController` that all generated decorators inherit from, run:
> 
> ```
> rails generate draper:install
>````

I don't think the additional comment about them inheriting from `Draper::Decorator` otherwise is necessary given that the paragraph above it says that all decorators inherit from `Draper::Decorator`.

I could also see this documentation living in the "Installation" section of the README at some point. I think we should leave it here for now, but just something to think about.

spec/generators/install/install_generator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+describe Draper::Generators::InstallGenerator do
+  destination File.expand_path('../tmp', __FILE__)
+
+  before { prepare_destination }
+  after(:all) { FileUtils.rm_rf destination_root }
+
+  describe 'the application decorator' do
+    subject { file('app/decorators/application_decorator.rb') }
+
+    describe 'naming' do
```
syguer: I think we doesn't need this block, does it?

lib/generators/draper/templates/application_decorator.rb
chuck-john: @codebycliff, I was basing this template off of the [decorator template](https://github.com/drapergem/draper/blob/master/lib/generators/rails/templates/decorator.rb).  Do you want the two to match?

lib/generators/draper/templates/application_decorator.rb
```diff
@@ -0,0 +1,8 @@
+class ApplicationDecorator < Draper::Decorator
```
chuck-john: I removed `delegate_all` from this template.  The belief is that certain issues and bad practices may stem from having this line in all decorators by default.  Implementers should be encouraged to whitelist their delegations to minimize this risk.

codebycliff: I agree completely. It's best not to assume functionality at the application level. I'd rather the user add this line themselves that way they are forced to understand what it does.

syguer: I agree your opinion too :)
However,  how do you think about leaving as comment to notice function of `delegate_all`? And additionally, if you agree with me, I want to add `include Draper::LazyHelpers`.

syguer:
I've added [a bit comment](https://github.com/drapergem/draper/pull/796#discussion_r109977938).
But great job. Patch looks good to me 👍

lib/generators/draper/templates/application_decorator.rb
codebycliff: That's an interesting idea. If we left in any commented out stuff, I'd want to include _all_ draper functionality that can be used (e.g. `decorates_finders`) with documentation explaining. I think I'd like to wait for this though for the time being unless you feel strongly about it.

PR #799, opened 2017-05-05
Clean view context when controller changes

barthez:
## Description
When we reassign controller it is the right thing to do to clear current view context stored in request store. Consider a case when view context was built before the request was made (usage of `h` helper in class level context). View context is built from default controller and has no information about the request. During first request `Draper::ViewContext.current.controller` will be different than `Draper::ViewContext.controller`.

## Testing

Example Rails app. Use `h.content_tag` in class context of some decorator, make this class be eagerly loaded.

Perform some request that will execute some decorator to refer to e.g. `request.env` values that are set up by some middleware (e.g. warden manager).

See that `env` is missing that key.

With that fix, view_context is regenerated when a controller is already available.

## To-Dos
- [x] tests
- [x] documentation

### DIFF
--- a/lib/draper/view_context.rb
+++ b/lib/draper/view_context.rb
@@ -20,8 +20,10 @@ module Draper
       RequestStore.store[:current_controller]
     end
 
-    # Sets the current controller.
+    # Sets the current controller. Clears view context when we are setting
+    # different controller.
     def self.controller=(controller)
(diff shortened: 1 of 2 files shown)


barthez:
Fix for failure can be found here: https://github.com/drapergem/draper/pull/800

PR #801, opened 2017-05-05
Revert "Optimization: Optimize CollectionDecorator Enumerable behavior"

codebycliff:
I'm reverting this pull request for the time being. It was just a performance optimization to begin with. It caused some issues as noted [in this thread](https://github.com/drapergem/draper/commit/f6a7b17a8acfa7633762d665ab218a6478d32a98). Furthermore, running the benchmarks again are returning inconsistent results. This time around, it's showing the performance being slightly better pre-patch. I'm not sure if this is just do to inconsistencies on my machine or what. Either way, I think this is the safest bet until we can resolve the issues noted above and provide convincing benchmarks.

### DIFF
--- a/lib/draper/collection_decorator.rb
+++ b/lib/draper/collection_decorator.rb
@@ -15,7 +15,7 @@ module Draper
     #   to each item's decorator.
     attr_accessor :context
 
-    array_methods = Array.instance_methods - Object.instance_methods - Enumerable.instance_methods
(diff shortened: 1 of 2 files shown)


clemens:
Too bad that this didn't work without issues.

The part with performance being better pre-patch seems odd to me since there are **obviously** fewer iterations. But since there are other issues anyway, I guess that doesn't matter much.

PR #806, opened 2017-05-10
Replace double quotes with single quotes

DmytroStepaniuk:
# Proposal request

## Description

I have noticed that some of string constants wrapped in double quotes, and some of them - in single quotes, so, my proposal is - use one approach to specify string constants and don't mix them.

## Things to discuss 

Double quotes vs single - which approach has better performance?

Example: 

https://github.com/vigetlabs/ruby-string-showdown

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,15 +1,15 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 gemspec
 
 platforms :ruby do
-  gem "sqlite3"
+  gem 'sqlite3'
 end
 
 platforms :jruby do
-  gem "minitest"
-  gem "activerecord-jdbcsqlite3-adapter"
+  gem 'minitest'
+  gem 'activerecord-jdbcsqlite3-adapter'
 end
 
-gem "rails", "~> 5.0"
-gem "mongoid", github: "mongodb/mongoid"
+gem 'rails', '~> 5.0'
+gem 'mongoid', github: 'mongodb/mongoid'
--- a/Guardfile
+++ b/Guardfile
@@ -10,17 +10,17 @@ end
 rspec_guard spec_paths: %w{spec/draper spec/generators} do
   watch(%r{^spec/.+_spec\.rb$})
   watch(%r{^lib/(.+)\.rb$})     { |m| "spec/#{m[1]}_spec.rb" }
-  watch('spec/spec_helper.rb')  { "spec" }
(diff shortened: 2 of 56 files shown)


DmytroStepaniuk:
Detected problems with newest minitest 5.10.2  (was 5.10.1)

codebycliff:
Looks like rails fixed the bug [here](https://github.com/rails/rails/pull/29022) and it made it into the most recent release. So the builds are passing again 🎉 .

As far as the single quotes vs double quotes:

Single quotes are my preference, as well as the recommendation from the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) (hence Rubocop). I added Rubocop to the project a bit ago, but added an exception for string literals since there were so many. If you want to remove that configuration from `.rubocop.yml` and run rubocop to double check all usages have been fixed, I'd be happy to merge this. Thanks!

Tonkpils:
> as well as the recommendation from the Ruby Style Guide

Actually, they use the single quotes in the style guide but they don't recommend one or the other.

From https://github.com/bbatsov/ruby-style-guide#strings

> Adopt a consistent string literal quoting style. There are two popular styles in the Ruby community, both of which are considered good

As far as performance is concerned:
https://stackoverflow.com/questions/1836467/is-there-a-performance-gain-in-using-single-quotes-vs-double-quotes-in-ruby/1836838#1836838

The performance difference is negligible pretty much

DmytroStepaniuk:
Okay, updated,  `rubocop --only Style/StringLiterals` shows no Style/StringLiterals warnings!

codebycliff:
Sorry it's taking me so long to get around to this. Can you also remove the configuration from `.rubocop.yml` so that this change makes it into the build?

codebycliff:
@DmytroStepaniuk @tibbon Are either of you interested in getting this pull request up to date? If not, I'll probably close it and take care of it myself at a later date.

DmytroStepaniuk:
> @DmytroStepaniuk @tibbon Are either of you interested in getting this pull request up to date? If not, I'll probably close it and take care of it myself at a later date.

Yep, go on. Unfortunately it has conflicts yet, but I have no free time to dig inside 🤷‍♂️

U can still fork my fork to make it work 😎

**P.S.** Lemme know if U need some help with that ℹ️

codebycliff:
Sounds good. I'm going to close this for now and take a more incremental approach. Thanks for the initial work and sorry it didn't end up making it through.

PR #816, opened 2017-07-15
Refactor generator for TestUnit, RSpec

yhirano55:
Refactored generators:

1. Add `check_class_collision suffix: "DecoratorTest"` to generator for TestUnit
2. Nested namespace `Generators` same as originals ([test_unit](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/test_unit.rb#L4), [rspec](https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec.rb#L9)]

I've checked running on rails application for sandbox. No problems.

### DIFF
--- a/lib/generators/draper/install_generator.rb
+++ b/lib/generators/draper/install_generator.rb
@@ -1,7 +1,7 @@
 module Draper
   module Generators
     class InstallGenerator < Rails::Generators::Base
-      source_root File.expand_path('../templates', __FILE__)
+      source_root File.expand_path("templates", __dir__)
 
       desc 'Creates an ApplicationDecorator, if none exists.'
 
--- a/lib/generators/mini_test/decorator_generator.rb
+++ b/lib/generators/mini_test/decorator_generator.rb
@@ -4,7 +4,7 @@ module MiniTest
   module Generators
     class DecoratorGenerator < Base
       def self.source_root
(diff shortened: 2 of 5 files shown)


lib/generators/draper/install_generator.rb
```diff
@@ -1,7 +1,7 @@
 module Draper
   module Generators
     class InstallGenerator < Rails::Generators::Base
-      source_root File.expand_path('../templates', __FILE__)
+      source_root File.expand_path("templates", __dir__)
```
codebycliff: Is there a reason for this change? I understand that it will do the same thing. However, when generating a generator, Rails still outputs the original. I'm curious what the inspiration was or why you prefer this?

yhirano55: I really appreciate for your comment. My opinion is very simple. It'll better that be shorter than before. And It'll be easier to understand than before, because directory layers that maintainers imagine are less than before. Of cause if you don't like this change, I'll follow your opinion :)

codebycliff: I agree that it is easier to understand conceptually. My only concern is that it differs from the Rails convention. I'd be curious what the Rails core team would say to a pull request to change this in the Rails codebase. Either way, I am fine with. Thanks for the changes!

PR #820, opened 2017-09-05
Let decorator_class infer anonymous class' decorator from superclass

masolin:
## Description
Since Rails 5, draper gem gets some issues of raising `NameError` on inferring decorator class (https://github.com/drapergem/draper/issues/773). The previous PR (https://github.com/drapergem/draper/pull/795) fixes this issue when using `decorate_collection`, but this issue can also happen when using `decorate`, and we encountered this issue on our production code when upgrading to Rails 5.

I tried to replace `constantize` with `safe_constantize` in this PR, so we don't need to rescue and check if the `NameError` was caused by draper itself anymore. It should be safer than previous approach.

Thanks.

## References
* [GitHub Issue #773 ](https://github.com/drapergem/draper/issues/773)
* [GitHub Pull Request #795 ](https://github.com/drapergem/draper/pull/795)

### DIFF
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -222,11 +222,9 @@ module Draper
     # @return [Class] the class created by {decorate_collection}.
     def self.collection_decorator_class
       name = collection_decorator_name
-      name_constant = name.safe_constantize
+      name_constant = name && name.safe_constantize
 
       name_constant || Draper::CollectionDecorator
-    rescue NameError
-      Draper::CollectionDecorator
     end
 
     private
@@ -241,25 +239,23 @@ module Draper
     end
 
     def self.object_class_name
-      raise NameError if name.nil? || name.demodulize !~ /.+Decorator$/
+      return nil if name.nil? || name.demodulize !~ /.+Decorator$/
       name.chomp("Decorator")
     end
 
     def self.inferred_object_class
       name = object_class_name
-      name_constant = name.safe_constantize
+      name_constant = name && name.safe_constantize
       raise Draper::UninferrableObjectError.new(self) if name_constant.nil?
 
       name_constant
-    rescue NameError => error
-      raise if name && !error.missing_name?(name)
-      raise Draper::UninferrableObjectError.new(self)
     end
 
     def self.collection_decorator_name
-      plural = object_class_name.pluralize
-      raise NameError if plural == object_class_name
-      "#{plural}Decorator"
+      singular = object_class_name
+      plural = singular && singular.pluralize
+
+      plural == singular ? nil : "#{plural}Decorator"
     end
 
     def handle_multiple_decoration(options)
--- a/lib/draper/decorator.rb
+++ b/lib/draper/decorator.rb
@@ -246,16 +246,16 @@ module Draper
     def self.inferred_object_class
       name = object_class_name
       name_constant = name && name.safe_constantize
-      raise Draper::UninferrableObjectError.new(self) if name_constant.nil?
+      return name_constant unless name_constant.nil?
 
-      name_constant
+      raise Draper::UninferrableObjectError.new(self)
     end
 
     def self.collection_decorator_name
       singular = object_class_name
       plural = singular && singular.pluralize
 
-      plural == singular ? nil : "#{plural}Decorator"
+      "#{plural}Decorator" unless plural == singular
     end
 
     def handle_multiple_decoration(options)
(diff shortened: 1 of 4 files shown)


codebycliff:
@masolin Thanks for the contribution! I think this change makes sense, but give me a little bit to think through any unintended side affects. Also, if we do move forward with this, I think it would make sense to update all the other usages of `constantize` to use `safe_constantize` as well.

@steveklabnik @jcasimir This change makes sense to me, but wanted to double check with one of you guys in case there is some historical knowledge I'm missing.

steveklabnik:
It's been too long; I don't remember any details here, sorry!

masolin:
@codebycliff Please let me know if you think that it's ok to use `safe_constantize`. I will change the other part as well. Thanks.

codebycliff:
@masolin Thanks for your patience on this. I'm good with this change. If you can get the other uses of `constantize` updated to use `safe_constantize` we can get this merged. Thanks again!

masolin:
Hi @codebycliff 
I replaced `constantize` with `safe_constantize` on most of the parts except for `self.decorates`, because it's a public method and makes sense to return `NameError`.
In addition, I also revised a private method, `object_class_name`, to return `nil` rather than `NameError`, so we don't need to rescue and check if  `NameError` is from `draper` gem anymore.

Please help to review it. Thanks.

lib/draper/decorator.rb
```diff
@@ -222,9 +222,9 @@ def attributes
     # @return [Class] the class created by {decorate_collection}.
     def self.collection_decorator_class
       name = collection_decorator_name
-      name.constantize
-    rescue NameError
-      Draper::CollectionDecorator
+      name_constant = name && name.safe_constantize
+
+      name_constant || Draper::CollectionDecorator
```
codebycliff: @masolin How do you feel about using the safe navigation operator here? I believe it reads better and is more succint. This could become:

```ruby
name&.safe_constantize || Draper::CollectionDecorator
```

lib/draper/decorator.rb
```diff
@@ -239,22 +239,23 @@ def self.alias_object_to_object_class_name
...
     def self.inferred_object_class
       name = object_class_name
-      name.constantize
-    rescue NameError => error
-      raise if name && !error.missing_name?(name)
-      raise Draper::UninferrableObjectError.new(self)
+      name_constant = name && name.safe_constantize
+      raise Draper::UninferrableObjectError.new(self) if name_constant.nil?
+
+      name_constant
```
codebycliff: In `Decoratable#decorator_class` above, you are returning the constant unless it's nil and raise at the end. Here you are doing the exact opposite. For consistency purposes, I think it would make sense to follow the logic from above here as well:

```ruby
name_constant = name&.safe_constantize
return name_constant unless name_constant.nil?
raise Draper::UninferrableObjectError.new(self)
```

lib/draper/decorator.rb
```diff
@@ -239,22 +239,23 @@ def self.alias_object_to_object_class_name
...
     end
 
     def self.collection_decorator_name
-      plural = object_class_name.pluralize
-      raise NameError if plural == object_class_name
-      "#{plural}Decorator"
+      singular = object_class_name
+      plural = singular && singular.pluralize
+
+      plural == singular ? nil : "#{plural}Decorator"
```
codebycliff: While the ternary statement works just fine, returning `nil` is the default behavior. So it might read better to say:

```ruby
singular = object_class_name
plural = singular&.pluralize
"#{plural}Decorator" unless plural == singular
```

What do you think?

lib/draper/decorator.rb
masolin: If we still want to support Ruby 2.2 and 2.3, then we shouldn't use safe navigation operator, right?

codebycliff: Haha, yes good catch. Sorry about that.

lib/draper/decorator.rb
masolin: I feel that the ternary statement can make me more easily realize the 2 possible return types: `nil` and `"#{plural}Decorator"`, but I don't have strong opinion on it, so it's fine to change

```ruby
"#{plural}Decorator" unless plural == singular
```

codebycliff: My opinion isn't strong either. I see the advantages of the ternary as well. Completely up to you.

masolin:
Thanks a lot. 😄

codebycliff:
This has been released to RubyGems.

Issue #819, opened 2017-08-18
Misleading Error Message: Could not infer a decorator for ActiveRecord::Base.


pascalbetz:
Since `decorator_class` recursevly walks up the class hierarchy to find a decorator the error is misleading and does not show the real class name. As an example:

```ruby
thing = MyCoolProject::CoolThing.new(...)
thing.decorate
```

will raise following error when decorator can not be found

> Could not infer a decorator for ActiveRecord::Base.

instead of

> Could not infer a decorator for MyCoolProject::CoolThing

Not a big issue as I usually now what I'm decorating...but I'm a big fan of helpful error messages:-)

I can provide a fix if this is of interest.

syguer:
Related https://github.com/drapergem/draper/issues/767

swelther:
I get this misleading error every now and then. Really annoying, would be happy for a fix :blush:

pascalbetz:
@swelther I created a PR, feedback welcome!

codebycliff:
Closing. This was fixed in #821.

codebycliff:
This has been released to RubyGems.

masciugo:
I had to add to point to github
`gem 'draper', github: 'drapergem/draper'`

to have it working

vnhnhm:
As stated in the docs, it might help if nothing else works:

```ruby
@widget = ProductDecorator.new(Widget.first)
# or, equivalently
@widget = ProductDecorator.decorate(Widget.first)
```

PR #821, opened 2017-09-13
When decoration fails, show original class

pascalbetz:
## Description
Thanks for the work on Draper to all the maintainers 👍

This provides a fix for #819 

When decorating an object without decorator, then the last class which was checked for a decorator is contained in the error message. In case of Rails this is most often `ActiveRecord::Base`.
This is a bit misleading and it would be better if the class which was actually tried to decorate is shown in the error message.
This is caused by the recursive nature of the `decorate` method which moves up  the hierarchy when no decorator class found. And raises when it can not move further up.

Before:

```ruby
Something::Bar.decorate
Draper::UninferrableDecoratorError: Could not infer a decorator for ActiveRecord::Base.
```

After:

```ruby
Something::Bar.decorate
Draper::UninferrableDecoratorError: Could not infer a decorator for Something::Bar.

```
## Testing
Outline steps to test your changes.

* Run the specs, tests have been added

Alternatively:
* on a subclass of AR, which does not have a decorator, call #decorate
* Check that the name of the subclass is contained in the error message and not `ActiveRecord::Base`

(The fix is not restricted to AR)

## References
* https://github.com/drapergem/draper/issues/819

### DIFF
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -69,16 +69,16 @@ module Draper
       # `Product` maps to `ProductDecorator`).
       #
       # @return [Class] the inferred decorator class.
-      def decorator_class
+      def decorator_class(called_on = self)
         prefix = respond_to?(:model_name) ? model_name : name
         decorator_name = "#{prefix}Decorator"
         decorator_name_constant = decorator_name.safe_constantize
         return decorator_name_constant unless decorator_name_constant.nil?
 
         if superclass.respond_to?(:decorator_class)
-          superclass.decorator_class
+          superclass.decorator_class(called_on)
         else
-          raise Draper::UninferrableDecoratorError.new(self)
+          raise Draper::UninferrableDecoratorError.new(called_on)
         end
       end
 
--- a/spec/draper/decoratable_spec.rb
+++ b/spec/draper/decoratable_spec.rb
@@ -73,6 +73,16 @@ module Draper
         expect(Product).to receive(:decorator_class).and_return(:some_decorator)
         expect(product.decorator_class).to be :some_decorator
       end
+
+      it "specifies the class that #decorator_class was first called on (superclass)" do
+        person = Person.new
+        expect { person.decorator_class }.to raise_error(Draper::UninferrableDecoratorError, 'Could not infer a decorator for Person.')
+      end
+
+      it "specifies the class that #decorator_class was first called on (subclass)" do
+        child = Child.new
+        expect { child.decorator_class }.to raise_error(Draper::UninferrableDecoratorError, 'Could not infer a decorator for Child.')
+      end
     end
 
     describe "#==" do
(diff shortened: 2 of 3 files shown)


lib/draper/decoratable.rb
```diff
@@ -69,16 +69,16 @@ def decorator_class?
       # `Product` maps to `ProductDecorator`).
       #
       # @return [Class] the inferred decorator class.
-      def decorator_class
+      def decorator_class(called_on = self)
```
codebycliff: @pascalbetz Thanks for the contribution! I am not crazy about a public(ish) method taking a parameter that is only relevant to draper's internals. However, I can't think of a better way to accomplish this and it definitely solves a longstanding (and very annoying) problem.

Can you `rebase` with master before I merge? There has been some changes to this method in master as well and want to make sure the build passes. After that, I think we are good to go.

pascalbetz: @codebycliff I understand.
This was the easiest way to solve the problem... Alternatively we could change from recursion to iteration, this would get rid of the parameter. WDYT?

codebycliff: @pascalbetz It looks like however you merged, it brought in `master` changes to this pull request... Can you try again with `rebase`? 

As far as the solution goes, I'm not sure. I'm not super opposed to this solution, however, I am interested in what the iteration solution looks like. Especially now that `decorator_class` returns  `nil` if can't be found. It might make the iteration easier.

pascalbetz:
@codebycliff  sorry for messing up the rebase (thought I can easily do this while eating, watching a movie and solve some SQL stuff I'm working on. Looks like I'm not as smart as I thought ;-) ). Will clean this up when you decide to merge the recursive solution instead of the iterative solution at https://github.com/drapergem/draper/pull/823

codebycliff:
@pascalbetz I think we'll go with the recursive solution for now. It's much simpler to read and understand. Plus, `decorator_class` isn't truly public API as it's not documented. I might want to re-visit the iterative solution at some point, but let's move forward with this for now. If you can clean up the commit history we can get this merged.

pascalbetz:
@codebycliff PR should be cleaner now.


Here is the iterative solution for documentation purpose:

```
def decorator_class
        current = self
        decorator_name_constant = nil

        while !decorator_name_constant && current do
          prefix = current.respond_to?(:model_name) ? current.model_name : current.name
          decorator_name = "#{prefix}Decorator"
          decorator_name_constant = decorator_name.safe_constantize
          if !decorator_name_constant && current.superclass.respond_to?(:decorator_class)
            current = current.superclass
          else
            current = nil
          end
        end

        if !decorator_name_constant
          raise Draper::UninferrableDecoratorError.new(self)
        end

        decorator_name_constant
end
```

codebycliff:
@pascalbetz Thanks again! Sorry it took me so long to get it merged.

pascalbetz:
No problem at all. Thanks for your work on Draper. I use it in various projects.

PR #836, opened 2018-10-26
Add missing verb to README

matildasmeds:
Just adds missing "be" verb to README


codebycliff:
I believe this change produces less grammatically correct english than was already there. If you wanted to add the verb, you would need to say "can either be switch_ed_ based..." and even then I don't know if it adds any value.

PR #838, opened 2018-12-26
Testing against latest 2.5 version

lucianosousa:
## Description
Testing against latest 2.5 version

* Why was this change required?
Since 2.6 was released, ruby 2.5 is not ruby-head anymore


codebycliff:
Thanks for this! The help is much appreciated. Sorry for the late reply.

Issue #814, opened 2017-07-11
Should automatically clear view_context before/after each spec


zinkkrysty:
So that it doesn't leak from a mailer spec to a controller spec and viceversa, since the helpers are different in those contexts.

This probably causes all these issues:
#523 #813 #655 #580

We need a solution that addresses these issues.

codebycliff:
Thanks for reporting this issue. I have checked the issues you referenced and it does appear that they are all related. I will start looking into it, but I may need to gather some context from some of the original authors as to why we aren't doing this.

Furthermore, it looks like we are clearing the view context `before` each spec (https://github.com/drapergem/draper/blob/master/lib/draper/test/rspec_integration.rb). In one of the referenced issues, you said adding an `after` hook to clear the view context fixed it. I question why it is needed in both `before` and `after` hooks.

mobject:
I am having same issue when running several mailers and the Context was not cleared out

codebycliff:
@tibbon I was able to contact one of the original authors, but they couldn't remember enough context to be of any help. I have dug through all the related issues and put together a rough overview and timeline of the issue.

## Overview
While this issue seems to exist for both Minitest and RSpec, it seems to be more common with RSpec. I just merged a [pull request](https://github.com/drapergem/draper/pull/839) for the Minitest side of things that hopefully will clear up any remaining issues for Minitest. For the rest of this comment / issue, I'll focus on RSpec.

## Timeline
1. The first attempt to fix this issue was in 2013 and is [located here](https://github.com/drapergem/draper/commit/3d07cb387b1cae6f97897dfb85512e30f5e888e9). It started out as an `after(:each)` for `decorator`, `mailer` and `controller` specs.
1. A couple months later, [this issue](https://github.com/drapergem/draper/issues/523) was opened. A lot of great discussion came out of that issue.
    - The initial conversation lead to it being changed to a `before(:each)` [here](https://github.com/drapergem/draper/pull/547/files). This was still in 2013.
    - That issue continued to get conversation for a couple years, eventually leading to [a proposed solution](https://github.com/drapergem/draper/issues/523#issuecomment-314414888) of using **both** `before(:each)` and `after(:each)`.

## Proposed solution
[...]

Issue #813, opened 2017-07-06
Decorator can not access controller helper in test


swelther:
Today I noticed a weird behaviour during my request spec:

The controller defines a helper method which is called from the decorator. But during the test this helper is not available.

Please see this gist with a sample app that reproduces the problem, using Ruby 2.3.3:

https://gist.github.com/swelther/47c5ffb707ca1145ca87d1dfb8aa99fa

My real world app uses RSpec instead of Minitest, but the result is the same.

Shouldn't it be possible to access helpers defined in the controller? What can I do to fix this?

codebycliff:
Thanks for reporting this issue. It looks like it has been around for a while. It appears to have to do with the `ViewContext` not being cleared after each spec. I'll look into this and get back to you.

Alexander-Senko:
1. I've got the same result after modernizing the gist for `master` and latest Rails.

2. After a lot of debugging and digging in, it appeared that Draper initializers don't get run for some reason. Thus, adding one to the gist fixes the issue:

```ruby
ActiveSupport.on_load :action_controller do
  Draper.setup_action_controller self
end
```

PR #839, opened 2019-01-04
Move view context clear to before_setup (minitest)

showaltb:
We are using Draper with Minitest and experiencing the "leakage" of mailer view contexts into decorator tests. See #814 and related.

Draper does attempt to prevent the leakage by clearing the view context in a `teardown` hook, but this isn't effective. We are still seeing the problem.

This PR moves the clearing from `teardown` to `before_setup`, which properly resolves the leakage.

### DIFF
--- a/lib/draper/test_case.rb
+++ b/lib/draper/test_case.rb
@@ -3,9 +3,9 @@ module Draper
 
   class TestCase < ::ActiveSupport::TestCase
     module ViewContextTeardown
-      def teardown
-        super
+      def before_setup
         Draper::ViewContext.clear!
+        super
       end
     end
 


codebycliff:
Thanks for this! I know this view context "leakage" has been a major pain point for a long time. The majority of the issues I have seen have been with RSpec, but I did [find one for minitest](https://github.com/drapergem/draper/issues/813). That issue includes a gist that reproduces the issue. Interestingly enough, that gist still fails using this branch [(see updated gist with your branch here](https://gist.github.com/codebycliff/840c6a4a8a38f18e9e9b72a845c17a24)) I'm not _entirely positive_ that issue is the _exact_ same as the view context leaking in between tests as that gist only has one test in it.

According to [the docs (http://docs.seattlerb.org/minitest/Minitest/Test/LifecycleHooks.html), `before_setup` "is meant for libraries to extend minitest". Because of that, I agree that we should be using `before_setup` over `teardown` regardless. 

How did you go about confirming this fixes the issue? If at all possible, it would be nice to have some sort of test or reassurance that this works. If not, I'm fine merging this and seeing if it fixes the issue as I don't think there will be any negative consequences.

showaltb:
I think the issue in that gem is different. The problem there is that the helper method defined in the Books controller is not accessible in the decorator. It sounds like a lifecycle issue where Draper is grabbing the view context before the controller-defined helper methods have been injected.

In my case, I have an app that contains both mailer tests and decorator tests, where the decorator access path helpers (e.g. `widgets_path`). If a mailer test was run immediately before a decorator test, the mailer context would leak into the decorator test and the latter would fail with `unknown method widgets_path`, and suggest `widgets_url` as an alternative. Moving the clearing to `before_setup` resolved this.

I'll see if I can work up a simple example like the gist that illustrates the problem and demonstrates the fix.

showaltb:
I've spent most of the day and unfortunately I cannot come up with a trivial example that fails. The `teardown` hook seems to do the job, except for my specific app.

I doubt that moving to `before_setup` will hurt anything, but I can't prove that it fixes anything except for my specific situation. So I understand if you don't want to accept this PR.

showaltb:
Thanks @codebycliff!

Issue #812, opened 2017-06-22
Consider implement #preload in Draper::CollectionDecorator


teohm:
We're using [datagrid gem](https://github.com/bogdan/datagrid) together with draper decorators. 

They works very well so far, until recently [datagrid introduces preload feature](https://github.com/bogdan/datagrid/commit/3c756c383ed8ad8ddfd4e74cd502df3242f4bf57), where it actively preloads associations when rendering a ActiveRecord resource.

To fix this breaking change, we have to implement `#preload` in our `Draper::CollectionDecorator` subclasses:

```
class MyCollectionDecorator < Draper::CollectionDecorator
  def preload(*args)
    object.preload(*args).decorate
  end
end
```

I wonder if it makes sense to move this `#preload` implementation into `Draper::CollectionDecorator`? 

What do you all think?

codebycliff:
Thanks for the suggestion!

I've seen a few issues related to this functionality. Let me look into those and get back to you. At first glance I don't see a problem with adding this functionality, but I'll want to get a little more context before I give you a full answer.

Alexander-Senko:
> I wonder if it makes sense to move this `#preload` implementation into `Draper::CollectionDecorator`?

IMO, it doesn't as `Draper::CollectionDecorator` may decorate collections not backed by a DB. It's not meant to be used solely with Active Record, as I can see it.

Instead, there is [a proposal of `RelationDecorator`](https://github.com/drapergem/draper/pull/662) that could solve this.

Issue #702, opened 2015-10-20
undefined method `includes' for #<Draper::CollectionDecorator:0x007fa8f21c0ed8>


adsteel:
I am having a similar problem to [this issue](https://github.com/drapergem/draper/issues/683). After calling `delegate_all` in my delegator, and decorating that records' associations, I am unable to call ActiveRecord's `includes`.

``` ruby
class AuthorDecorator < Draper::Decorator
  delegate_all
  decorates_association :posts
  ...
end

class BlogDecorator < Draper::Decorator
  delegate_all
  ...
end

@author.decorate.posts.includes(:comments)

# raises NoMethodError
# undefined method `includes' for #<Draper::CollectionDecorator:0x007fa8f21c0ed8>
```

This is a simplified example, and calling `.decorate` after the `.includes` isn't an option.

lucianosousa:
can you try:
`AuthorDecorator.new(@author).posts.includes(:comments)`

and see what is happening?

adsteel:
Sure. That returns the same error. Thanks for your quick response!

Mohit-Pawar:
hi @adsteel can you try this.

class AuthorDecorator < Draper::Decorator
  delegate_all
  decorates_association :posts, with: PostsDecorator
  ...
end
class PostsDecorator < Draper::CollectionDecorator
	delegate :includes
end
or
@author.decorate.posts.object.includes(:comments) 
this option is not ideal

adsteel:
@Mohit-Pawar Thanks, I had put this feature on hold, but I'll try that when it starts up again. Do you know why includes isn't automatically delegated with `delegate_all` ?

Mohit-Pawar:
@adsteel 
The main reason why includes isn't automatically delegated with delegate_all is we are 
delegating all methods for AuthorDecorator not Draper::CollectionDecorator
1).
@author.decorate.posts will return Draper::CollectionDecorator Object
but Draper::CollectionDecorator delegates only following methods.

array_methods = Array.instance_methods - Object.instance_methods
delegate :==, :as_json, *array_methods, to: :decorated_collection
see source draper/lib/draper/collection_decorator.rb line 18

that is why @author.decorate.posts.first will work and @author.decorate.posts.includes(:comments) will not

But,
Now By using below code.
decorates_association :posts, with: PostsDecorator

@author.decorate.posts will return Draper::PostsDecorator Object
and we are explicitly delegating includes method in Draper::PostsDecorator

2).
[...]

PR #845, opened 2019-02-20
feat: Include ORM associations in CollectionDecorator

brunohkbx:
## Description
Include all QueryMethods from the ORM in CollectionDecorator. The default adapter is :active_record

* Why was this change required?
It was necessary to delegate or define a method of the ORM which you are using in your decorator to make an instance of CollectionDecorator able to call it.
* Is there something you aren't happy with or that needs extra attention?
In order to support other ORM associations, we'll need to write a method `allowed?` for each strategy at `lib/draper/query_methods/load_strategy.rb`

## Testing
1. Create a decorator for the model and its association
```ruby
class OrderHistoryDecorator < Draper::Decorator
  delegate_all
end

class OrderDecorator < Draper::Decorator
  delegate_all

  decorates_association :order_histories, with: OrderHistoryDecorator
end
```

2. Call any query method in the decorated instance
```ruby
pry(main)> Order.last.decorate.order_histories.includes(:user)
```

## References
* Issue #702 
* Issue #812


spec/draper/query_methods_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper'
+
+module Draper
+  describe QueryMethods do
+    describe "#method_missing" do
```
fladson: I think you forgot the single quotes on this one :nail_care:

spec/draper/query_methods/load_strategy_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
+require 'active_record'
+
+module Draper
+  module QueryMethods
+    describe LoadStrategy do
+      describe "#new" do
```
fladson: same here, since you are using single quotes on the other tests

spec/draper/query_methods_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      let(:instance) { Struct.new(:query_methods){include Draper::QueryMethods}.new(double) }
+      let(:fake_strategy) { instance_double(QueryMethods::LoadStrategy) }
+
+      before { allow(QueryMethods::LoadStrategy).to receive(:new).and_return(fake_strategy) }
+
+      context 'When strategy allows instance to call the method' do
+        before { allow(fake_strategy).to receive(:allowed?).with(:foo).and_return(true) }
+
+        it 'calls the method' do
+          allow(instance).to receive(:foo)
```
fladson: this can't be on the before?

lib/draper/query_methods.rb
```diff
@@ -0,0 +1,19 @@
+require 'draper/query_methods/load_strategy'
```
fladson: I think you can use `require_relative` here

spec/draper/configuration_spec.rb
```diff
@@ -6,20 +6,40 @@ module Draper
...
 
-    it 'defaults default_controller to ApplicationController' do
-      expect(Draper.default_controller).to be ApplicationController
-    end
+    describe '#default_controller' do
+      it 'defaults default_controller to ApplicationController' do
+        expect(Draper.default_controller).to be ApplicationController
+      end
+
+      it 'allows customizing default_controller through configure' do
```
wenderjean: Did you want to write `allows customizing default_controller through configuration`?

wenderjean: Or even, `through Draper#configure`, `through configure method`

spec/draper/configuration_spec.rb
```diff
@@ -6,20 +6,40 @@ module Draper
...
+
+        Draper.configure do |config|
+          config.default_controller = CustomController
+        end
+
+        expect(Draper.default_controller).to be CustomController
 
-    it 'allows customizing default_controller through configure' do
-      default = Draper.default_controller
+        Draper.default_controller = default
```
wenderjean: What you think to put that line in an `after` block?
```ruby
after { Draper.default_controller = Draper.default_controller } 
```

spec/draper/configuration_spec.rb
```diff
@@ -6,20 +6,40 @@ module Draper
...
+      end
+    end
 
-      Draper.configure do |config|
-        config.default_controller = CustomController
+    describe '#default_query_methods_adapter' do
```
wenderjean: What you think to use `context` on the different scenarios, something like:
```ruby
describe '#default_query_methods_adapter' do
  subject { Draper.default_query_methods_adapter }

  context 'when there is no custom adapter' do
    it { is_expected.to eq(:active_record) }
  end

  context 'when using a custom adapter' do
    before do
      Draper.configure do |config|
        config.default_query_methods_adapter = :mongo
      end
    end

    it { is_expected.to eq(:mongo) }
  end
end
```

spec/draper/query_methods/load_strategy_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        subject { described_class.new(:active_record) }
+
+        it { is_expected.to be_an_instance_of(LoadStrategy::ActiveRecord) }
+      end
+    end
+
+    describe LoadStrategy::ActiveRecord do
+      describe "#allowed?" do
```
wenderjean: Same about single quotes here

spec/draper/query_methods_spec.rb
brunohkbx: Good catch, thanks.

spec/draper/query_methods_spec.rb
brunohkbx: I prefer to write the specs using the format: setup/exercise/verify.

lib/draper/query_methods.rb
brunohkbx: Good catch, thanks.

spec/draper/configuration_spec.rb
brunohkbx: I'm not sure :thinking:. I've just moved this old tests into a context, maybe these changes are out the scope of this pr?

spec/draper/configuration_spec.rb
brunohkbx: Thanks!

lib/draper/query_methods/load_strategy.rb
```diff
@@ -0,0 +1,25 @@
...
+        raise NotImplementedError
+      end
+
+      class ActiveRecord
+        def allowed?(method)
+          ::ActiveRecord::Relation::VALUE_METHODS.include? method
+        end
+      end
+
+      class Mongo
```
codebycliff: Should this technically be `Mongoid`, as it's the actual ORM we are referring to? If we ever wanted to add support for say [MongoMapper](http://mongomapper.com/), we'd need to be able to specify that.

lib/draper/query_methods/load_strategy.rb
```diff
@@ -0,0 +1,25 @@
+module Draper
+  module QueryMethods
+    module LoadStrategy
+      def self.new(name)
+        const_get(name.to_s.camelize).new
+      end
+
+      def allowed?(method)
```
codebycliff: Unless I'm missing something, I don't think this module method is needed.

lib/draper/query_methods/load_strategy.rb
brunohkbx: Yes, it's the referred ORM. Good catch.

brunohkbx:
Hey @codebycliff. Thanks for your feedback! :smiley: 
I've already pushed all changes and improved a test.
Let me know if everything is ok.

README.md
```diff
@@ -326,6 +326,17 @@ your `ArticleDecorator` and they'll return decorated objects:
...
 ```
 
+### Decorated Query Methods
+By default, Draper will decorate all [QueryMethods](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html)
+of ActiveRecord.
+If you're using another ORM, in order to support it, you can tell Draper to use a custom strategy:
+
+```ruby
+Draper.configure do |config|
+  config.default_query_methods_strategy = :mongo
```
codebycliff: We'll want to update the documentation to use `:mongoid` now that we renamed the adapter.

brunohkbx: Ops! I forgot to update the docs.

Issue #771, opened 2016-12-31
Name clash with private methods


fxn:
I solved a weird bug this week in an application. Draper was working correctly, but this application hit what could be a gotcha in `delegate_all`. I open this issue to explain it, and to ask if you'd like me to volunteer a doc patch. Let me explain.

The situation was:

```ruby
class JobDecorator < Draper::Decorator
  delegate :final, to: :location, prefix: true
  delegate_all

  private
  def location
    (model.location || Location.new).decorate
  end
end

class LocationDecorator < Draper::Decorator
  def link_to_google_maps
    ...
  end
end
```

The symptom was that you deployed, everything was fine, and after some minutes

```ruby
job.decorate.location_final.link_to_google_maps
```

crashed with a `NoMethodError`, saying `Location` objects do not respond to `link_to_google_maps` (I am making the names up). How on earth is `job.decorate.location_final` returning all of a sudden a `Location` instead of a `LocationDecorator`?

Problem is, `Job` instances have `location` in their public interface. When the app launches, `JobDecorator` has a private method called `location`, but as soon as someone invokes `job.decorate.location`, the `method_missing` is triggered because `location` is not a public method of the decorator, so it is (correctly) forwarded **and** redefined/cached in the decorator class. The private method is gone.

That was the bug, but once understood, the problem is in the implementation of the decorator, because `delegate_all` effectively (lazily) injects the public interface of `Job` into `JobDecorator`, so this is as if you were writing by hand both a public and a private method with the same name. And the last one wins.

My solution has been to adopt the convention of naming private methods with a leading underscore. That way you are forward-compatible with any public method in `Job`, since conventionally Ruby programmers do not put leading underscores in public methods.

Do you think it would be good to document this gotcha? Something like "Beware `delegate_all` injects the current and future public APIs of the model in here, make sure the name of any private methods in the decorator do not clash. A leading underscore is a forward-compatible convention to ensure that."

fxn:
Alternatively, if delegating to a method that clashes with an existing one in the decorator, perhaps you could raise. But gauging whether that is a good idea is more difficult for me, would need to have more perspective about usage of Draper out there. You are in a better position to choose.

seanlinsley:
I think it'd make sense to raise an error.

codebycliff:
Fixed by: https://github.com/drapergem/draper/pull/849

PR #849, opened 2019-03-15
fix: Don't delegate public methods overridden by a private in the decorator

brunohkbx:
Public methods defined on the object was being delegated to the decorator even though it was overridden by a private method.

## Testing
1. Create a simple PORO with a public method
```ruby
class Job
  def location
    Struct.new(:id) { }
  end
end
```

1. Create a decorator and override the public method with a private
```ruby
class JobDecorator < Draper::Decorator
  delegate_all

  private

  def location
    LocationDecorator.new(object.location)
  end
end
```

3. It should raise an exception when calling `.location` from the decorator
```ruby
[9] pry(main)> JobDecorator.decorate(Job.new).location
NoMethodError: private method `location' called for #<JobDecorator:0x0000559f113ba118>
```

## References
* Issue #771

### DIFF
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -665,7 +665,7 @@ module Draper
           expect(decorator.methods).not_to include :hello_world
         end
 
-        context 'When decorator overrides a public method defined on the object with a private' do
+        context 'when decorator overrides a public method defined on the object with a private' do
           let(:decorator_class) do
             Class.new(Decorator) do
               private
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -676,7 +676,7 @@ module Draper
             end
           end
 
-          let(:object) { Class.new{ def hello_world; end }.new }
+          let(:object) { Class.new { def hello_world; end }.new }
 
           it 'does not delegate the public method defined on the object' do
             decorator = decorator_class.new(object)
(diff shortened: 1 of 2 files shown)


spec/draper/decorator_spec.rb
```diff
@@ -664,6 +664,26 @@ def hello_world
           expect{decorator.hello_world}.to raise_error NoMethodError
           expect(decorator.methods).not_to include :hello_world
         end
+
+        context 'When decorator overrides a public method defined on the object with a private' do
```
codebycliff: I believe it's preferred to not capitalize these:

```suggestion
        context 'when decorator overrides a public method defined on the object with a private' do
```

spec/draper/decorator_spec.rb
```diff
@@ -664,6 +664,26 @@ def hello_world
...
+            Class.new(Decorator) do
+              private
+
+              def hello_world
+                'hello world'
+              end
+            end
+          end
+
+          let(:object) { Class.new{ def hello_world; end }.new }
```
codebycliff: Small spacing issue:

```suggestion
          let(:object) { Class.new { def hello_world; end }.new }
```

brunohkbx:
@codebycliff I was thinking the same. Could it be a breaking change? :thinking: 
What if we raise a deprecated warning instead of raising an error?

codebycliff:
@brunohkbx I think that's definitely an option. I'm also considering just bumping the major version when I release this. It's been a while since a major version has been released, so I don't think it would be that big of a deal.

springerigor:
@brunohkbx & @codebycliff I need your attention https://github.com/drapergem/draper/pull/875. Thanks in advance 🙂

PR #852, opened 2019-03-20
Fix some issues

brunohkbx:
## Description
Fix all issues related to alias and safe navigation operator

## References
* codeclimate [issues](https://codeclimate.com/github/drapergem/draper/issues)


andrerpbts:
Safe navigation doesn't work on ruby versions prior to 2.3. Are draper still providing support for older ruby versions?

brunohkbx:
Good catch @andrerpbts !! I think draper will not support any version prior to `2.3.5`. However, let's see @codebycliff thoughts on this.

PR #856, opened 2019-06-19
Don't require action_controller/test_case

Aesthetikx:
This was initially introduced in 738074f986507198b9a1862d74dee6a88aa3071b in order to allow draper to work in a rails console session, however this no longer seems necessary.

Requiring this file in non test environments has the consequence of breaking `ActionController::Live` as discussed in [rails issue 31200](https://github.com/rails/rails/issues/31200).

### DIFF
--- a/lib/draper/railtie.rb
+++ b/lib/draper/railtie.rb
@@ -47,7 +47,6 @@ module Draper
     end
 
     def initialize_view_context
-      require 'action_controller/test_case'
       Draper.default_controller.new.view_context
       Draper::ViewContext.build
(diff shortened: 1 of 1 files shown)


Aesthetikx:
Actually, I had not tested this with a decorator using the `h` helper, and indeed, this does seem to be a required `require`. Furthermore, this code is only used during rails console and rails runner, so it is not hit during a rails server, unless you have spring running and use a rails console, so this is not as much of a problem as I originally thought, so I am closing this PR.

PR #866, opened 2019-10-29
Add support for Rails 6.0 to `ViewContext::BuildStrategy`

y-yagi:
## Description

In Rails 6.0 the API for `ActionView::Base.new` is changing.
`ActionView::Base.new` now take arguments [1], otherwise you
will see the following deprecation warning:

```
DEPRECATION WARNING: ActionView::Base instances should be constructed with a lookup context, assignments, and a controller.
```

[1] https://github.com/rails/rails/commit/e17fe52e0e0ef0842b6c409e1110a862c4e005bc

`ViewContext::BuildStrategy::Fast` does not have lookup context, so use `empty`
method instead of `new` to build instance.

## Testing

I'm not sure test policy, so I did not update Rails version of test. But if need, I will update that to 6.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,11 @@ source "https://rubygems.org"
 gemspec
 
 platforms :ruby do
-  gem 'sqlite3'
+  if RUBY_VERSION >= "2.5.0"
+    gem 'sqlite3'
+  else
+    gem 'sqlite3', '~> 1.3.6'
+  end
 end
 
 platforms :jruby do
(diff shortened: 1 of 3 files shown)


codebycliff:
@GSto Thanks for the contribution! Getting the test suite dummy app upgraded has been needed for a while now. That would be killer and help verify this change if you were interested in doing that upgrade. If you don't think you'll have the time, we can probably get this merged still with hopes that we get around to upgrading the dummy application.

y-yagi:
Thanks for your response. OK. I upgraded the dummy application to Rails 6.0. 
But Rails 6.0 only supports  Ruby >= 2.5. So I keep using Rails 5.x for Ruby < 2.5.

y-yagi:
Thanks!

Issue #864, opened 2019-09-28
Draper::QueryMethods does not preserve context


synth:
**Given**: Creation of a Decorator::Collection instance with a context and calling `order` on that collection to pass database sort ordering through to the ActiveRecord::Association
**Expected**: The resulting collection should preserve the original context
**Actual**: It does not preserve the context. 

Example code: 

```ruby
class Posts < ActiveRecord::Base
end

class PostDecorator < Draper::Decorator
end

decorated_collection = PostDecorator.decorate_collection(Post.all, context: {user: "foo"})
decorated_collection.context == {user: "foo"} # true

decorated_collection = decorated_collection.order("created_at asc")
decorated_collection.context == {user: "foo"} # false, but should be true
```

I don't know if this is the proper long term fix for this, but this fix works in Draper::QueryMethods:
https://github.com/drapergem/draper/blob/dd32a4aa2cd16e2682d9e9c2486630ec53383442/lib/draper/query_methods.rb#L6-L10

to:

```ruby
object.send(method, *args, &block).decorate(context: self.context)
```

synth:
I would imagine that the context should be passed along anytime `decorate` is called and there is context object available in the current scope

synth:
Discovered a workaround until this is fixed which is to use a similar approach to adding Pagination support. In the application specific collection decorator class, you can add the `order` method and simply tack on the context after calling `super`

```ruby
class ApplicationDecorator < Draper::Decorator
  def self.collection_decorator_class
    PaginatingDecorator
  end
end

class PaginatingDecorator < Draper::CollectionDecorator
  # support for will_paginate
  delegate :current_page, :total_entries, :total_pages, :per_page, :offset

  def order(*args)
    collection = super
    collection.context = self.context
    collection
  end

  def paginate(*args)
    object.paginate(*args).decorate(context: self.context)
  end
end

```

codebycliff:
Fixed by: https://github.com/drapergem/draper/pull/868

PR #868, opened 2019-12-16
Add preservation of decorator options on QueryMethods

DouglasLutz:
## References
* [GitHub Issue #864](https://github.com/drapergem/draper/issues/864)

## Description
The context given to a Collection Decorator was being lost when an `ActiveRecord::QueryMethod` was being called on a `Draper::CollectionDecorator`.

E.g. 
```rb
decorated_collection = PostDecorator.decorate_collection(Post.all, context: {user: "foo"}).order("created_at asc")
decorated_collection.context #{}
decorated_collection.decorator_class #nil
```

This was happening because `method_missing` on `draper/lib/draper/query_methods.rb` is invoking ActiveRecord method with `send` and creating another instance of decorator to decorate the result.

```draper/lib/draper/query_methods.rb
   return super unless strategy.allowed? method 
  
   object.send(method, *args, &block).decorate 
 end 
```

The problem is that the object context and decorator_class were being lost on creation of another instance.
The solution to this was to pass the options alongside the `decorate` call, it creates the instance of `CollectionDecorator` with the old options.

## Testing
The new behaviour of the Collection Decorator is the following.

```rb
decorated_collection = PostDecorator.decorate_collection(Post.all, context: {user: "foo"}).order("created_at asc")
decorated_collection.context #{:user=>"foo"}
decorated_collection.decorator_class #PostDecorator
```

Also added a test to cover this alteration.

### DIFF
--- a/spec/draper/query_methods_spec.rb
+++ b/spec/draper/query_methods_spec.rb
@@ -11,7 +11,8 @@ module Draper
 
     describe '#method_missing' do
       let(:collection) { [ Post.new, Post.new ] }
-      let(:collection_decorator) { PostDecorator.decorate_collection(collection) }
+      let(:collection_context) { { user: 'foo' } }
+      let(:collection_decorator) { PostDecorator.decorate_collection(collection, context: collection_context) }
 
       context 'when strategy allows collection to call the method' do
         let(:results) { spy(:results) }
@@ -25,6 +26,12 @@ module Draper
           collection_decorator.some_query_method
 
           expect(results).to have_received(:decorate)
+        end
+
+        it 'calls the method on the collection and keeps the decoration options' do
+          collection_decorator.some_query_method
+
+          expect(results).to have_received(:decorate).with({ context: collection_context, with: PostDecorator })
         end
       end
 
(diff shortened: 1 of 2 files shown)


spec/draper/query_methods_spec.rb
```diff
@@ -11,7 +11,8 @@ module Draper
 
     describe '#method_missing' do
       let(:collection) { [ Post.new, Post.new ] }
-      let(:collection_decorator) { PostDecorator.decorate_collection(collection) }
+      let(:collection_context) { {user: 'foo'} }
```
brunohkbx:
```suggestion
      let(:collection_context) { { user: 'foo' } }
```

brunohkbx:
Hey, @DouglasLutz @synth. Good catch! :+1: 
Sorry, I didn't realize that on #845 

@codebycliff could you please take a look?

DouglasLutz:
@codebycliff I agree with you with the dependency of `decorator_class` and `context`, `Draper::QueryMethods` is modeled to `CollectionDecorator`.

I will apply the formatting suggestion and depending on your thoughts I can move the code there.

PR #871, opened 2020-01-07
CI: Update travis Ruby versions

Aesthetikx:
## Description
Towards a release supporting 2.7.x, this changes updates the travis builds:

- Ruby 2.3.x is EOL as of 2019-03-31,
- Ruby 2.7.0 is out as of 2019-12-15.

The other ruby minor versions seem up to date afaict.

## Testing
Waiting for a green traivs build.

## References
* [Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/)


Aesthetikx:
Looks like the remaining deprecation warnings may be from the mongoid dependency, but it is hard to tell with those stack traces.

.travis.yml
```diff
@@ -9,10 +9,10 @@ services:
   - mongodb
 
 rvm:
-  - 2.3.8
```
codebycliff: Since Ruby 2.3 hasn't reached EOL just yet, and we haven't officially decided to drop support for it, I feel like we should continue to test against it for now.

Aesthetikx: I think it did go EOL last year: https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/ ; but I agree with your sentiment and have added it back until a later date.

codebycliff: You are 💯 correct. Apparently, I don't know what year it is currently.

Issue #867, opened 2019-12-10
dont decorate nil class


bekicot:
I create a custom load strategy

```
module Draper
  module QueryMethods
    module LoadStrategy
      class ActiveRecordWithKlass
        def allowed?(method)
          @allowed ||=
            ::ActiveRecord::Relation::VALUE_METHODS +
            [:klass, :reflect_on_association]
          @allowed.include? method
        end
      end
    end
  end
end
```

It tries to decorate nil class in `lib/draper/query_methods.rb:7`
because 
```markdown
`scope`.reflect_on_association
``` 

is returning nil

It should return the value as is

PR #874, opened 2020-02-06
refactor: move module query-methods into collection-decorator

brunohkbx:
## Description
As we discussed on #868, it makes sense to move this module into CollectionDecorator since it depends upon `decorator_class` and `context`.


lib/draper/collection_decorator.rb
```diff
@@ -88,5 +98,10 @@ def item_decorator
         ->(item, options) { item.decorate(options) }
       end
     end
+
+    # Configures the strategy used to proxy the query methods, which defaults to `:active_record`.
+    def strategy
+      @strategy ||= LoadStrategy.new(Draper.default_query_methods_strategy)
```
codebycliff: Does the name `LoadStrategy` still make sense in this new context? Should it be something like `QueryMethodStrategy` or something more specific to query methods? I don't have a big opinion here, but it does seem odd.

n-rodriguez:
@brunohkbx can you please rebase your branch on master ? Thank you !

brunohkbx:
@n-rodriguez sure thing! I totally forgot this PR, sorry.

Alexander-Senko:
I think I'll have time to review this soon.

Meanwhile, @brunohkbx, could you please resolve the conflicts?

lib/draper/collection_decorator.rb
```diff
@@ -72,6 +71,17 @@ def replace(other)
       self
     end
 
+    # Proxies missing query methods to the source class if the strategy allows.
+    def method_missing(method, *args, &block)
+      return super unless strategy.allowed? method
```
Alexander-Senko, line 76: I'm new here, so please forgive my stupid question 🙏

Why do we need any _strategy_ in the first place? Isn't `object.respond_to?` enough?

Alexander-Senko (commented):
Decorated collection may be not backed by a DB. These strategies make no sense in that case (see #920).

lib/draper/collection_decorator.rb
Alexander-Senko: I see: it's for methods returning decoratable collections. It's not obvious for everyone, though (see #867).

Alexander-Senko, line 76:
```suggestion
      return super unless object.respond_to? method
```

lib/draper/collection_decorator.rb
```diff
@@ -72,6 +71,17 @@ def replace(other)
...
+
+      object.send(method, *args, &block).decorate(with: decorator_class, context: context)
```
Alexander-Senko, line 78:
```suggestion
      object.send(method, *args, &block).try(:decorate, with: decorator_class, context: context)
```

PR #875, opened 2020-02-10
fix: Check only object's private method when preventing delegation

springerigor:
## Description
https://github.com/drapergem/draper/pull/849 introduced a valid fix to prevent calling object's public method when a decorator overwrites it through a private one. There is one caveat, though. `private_methods` methods has `all` parameter set to `true` by default (https://ruby-doc.org/core-2.7.0/Object.html#method-i-private_methods). As a result all private methods are returned, instead of only the ones defined on the object.

This commit fixes the above issue by setting the `all` parameter to `false`.

## Testing
```ruby
class Job
  def fork
    Struct.new(:id) { }
  end
end

class JobDecorator < Draper::Decorator
  delegate_all
end
```

```ruby
JobDecorator.decorate(Job.new).fork
```
gives:
```
NoMethodError: private method `fork' called for #<JobDecorator:0x00007fb73ca977e8>
```

even though `fork` method is not defined within `JobDecorator`.
## References
* [GitHub Issue 771](https://github.com/drapergem/draper/issues/771)
* [GitHub Pull Request 849](https://github.com/drapergem/draper/pull/849)

### DIFF
--- a/lib/draper/automatic_delegation.rb
+++ b/lib/draper/automatic_delegation.rb
@@ -20,7 +20,7 @@ module Draper
 
     # @private
     def delegatable?(method)
-      return if private_methods.include?(method)
+      return if private_methods(false).include?(method)
 
       object.respond_to?(method)
     end
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -684,6 +684,18 @@ module Draper
             expect{ decorator.hello_world }.to raise_error NoMethodError
           end
         end
+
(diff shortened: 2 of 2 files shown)


sobrinho:
Maybe a spec?

brunohkbx:
@springerigor Good catch! I didn't notice that sorry.
Please, add a spec to doc this behavior here: `spec/draper/decorator_spec.rb`

springerigor:
I provided a spec. It is my first contribution to the gem, so please judge if it is good enough 🙂

springerigor:
Thanks 🙂

PR #759, opened 2016-09-02
add compatibility with latest rails 5.1

zhu1230:
Hi,
the latest rails 5.1 changed the "ActionController::TestRequest.create" method by adding one parameter.

https://github.com/rails/rails/pull/26092

so, with latest rails , I got:

```
actionpack/lib/action_controller/test_case.rb:38:in `create': wrong number of arguments (given 0, expected 1) (ArgumentError)
```

in order to keep updated with latest rails and compatible with current 5.0, I made this PR.

Hope it can help.

PR #888, opened 2020-09-10
Update BuildStrategy to be compatible with Ruby 2.7

paulsturgess:
When creating an `ActionController::TestRequest`, the params differ
based on the gem version of ActionPack.

Currently the check asserts the names of the params and expects
them to be `[:req, :controller_class]`. However on Ruby 2.7 this changes
to `[:rest, :args]`.

Rather than check for the names of the params, we can make this more
robust by checking if there are any params. As we know that old versions
of `TestRequest` didn't require any params to create an instance.

The previous workaround was introduced in: [GitHub Pull Request #759](https://github.com/drapergem/draper/pull/759)

### DIFF
--- a/lib/draper/view_context/build_strategy.rb
+++ b/lib/draper/view_context/build_strategy.rb
@@ -43,11 +43,11 @@ module Draper
         end
 
         def new_test_request(controller)
-          is_above_rails_5_1 ? ActionController::TestRequest.create(controller) : ActionController::TestRequest.create
-        end
-
-        def is_above_rails_5_1
-          ActionController::TestRequest.method(:create).parameters.first == [:req, :controller_class]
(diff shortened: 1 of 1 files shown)


paulsturgess:
I'm not actually sure how to test the backwards compatibility of this change. As I believe it would rely running the test suite against an old version of ActionPack/ActionController. I looked in the Travis config but couldn't see that kind of setup.

paulsturgess:
Interestingly I have now since been able to reproduce this on Ruby 2.6.3, so clearly it isn't a Ruby 2.7 thing 🤔

paulsturgess:
OK on further investigation, I tracked it down to a line our in our test suite which was stubbing `ActionController::TestRequest.create` 🤦

PR #892, opened 2020-10-22
Workaround https://github.com/rspec/rspec-mocks/issues/1348

ojab:
## Description
Right now it's not possible to do something like `expect(Draper::ViewContext.current).to receive(:render).and_call_original` since it's fails, workaround it.
Not sure if I should add test for that, sine it's a workaround that can be dropped if/when rspec-mocks will be fixed.

## Testing
* Specs are still passing, I guess.

## To-Dos
- [ ] tests
- [ ] documentation

## References
* [GitHub Issue ####](https://github.com/drapergem/draper/issues/####)
* [GitHub Pull Request ####](https://github.com/drapergem/draper/pull/####)


n-rodriguez:
@ojab can you please rebase your branch on master ?

ojab:
Sure, done.

lib/draper/helper_proxy.rb
```diff
@@ -10,7 +10,7 @@ def initialize(view_context)
     # Sends helper methods to the view context.
     ruby2_keywords def method_missing(method, *args, &block)
       self.class.define_proxy method
-      send(method, *args, &block)
+      view_context.send(method, *args, &block)
```
Alexander-Senko, line 13: Should we ever remove this? 🤔 Looks like `view_context` is in its place here.

Alexander-Senko (changes requested):
> Not sure if I should add test for that, since it's a workaround that can be dropped if/when rspec-mocks will be fixed.

I think we'd better have a test to check that nothing breaks when/if the workaround is removed.

Alexander-Senko:
I still can't get the rationale behind this PR, sorry. What scenario makes one to want this?

```ruby
expect(Draper::ViewContext.current).to receive(:render).and_call_original
```

I'm asking because the scenario or the expectation could be worked around instead of Draper itself.

PR #895, opened 2021-01-13
Fix decoration spec

tbuehlmann:
Draper::CollectionDecorator#{is_a?,kind_of?} triggers
decorating the object. When stubbing a method call
on a CollectionDecorator, RSpec itself will call #is_a?
on the object. See:

https://github.com/rspec/rspec-mocks/blob/ea4d8ea4532480500462cc7a633d3408656a4dcd/lib/rspec/mocks/proxy.rb#L38

That's why the object will always be decorated when
stubbing a method on it.

We can circumvent this issue by directly checking
the definition of @decorated_collection.

Edit:

This PR got bigger than intended, but it fixes more failing and flaky specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ end
 
 if RUBY_VERSION >= "2.5.0"
   gem "rails", "~> 6.0"
+  gem 'webrick'
 else
   gem "rails", "~> 5.0"
 end
--- a/draper.gemspec
+++ b/draper.gemspec
@@ -30,5 +30,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'capybara'
   s.add_development_dependency 'active_model_serializers', '>= 0.10'
   s.add_development_dependency 'rubocop'
-  s.add_development_dependency 'simplecov'
+  s.add_development_dependency 'simplecov', '0.17.1'
 end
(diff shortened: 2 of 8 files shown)


tbuehlmann:
Looks like the integration specs are flaky, gonna check.

tbuehlmann:
Righty, but I found a real flaky spec now. Could reproduce with `bundle exec rspec spec/generators --seed=53929`. Fixed now and hopefully CI will pass now.

tbuehlmann:
@mgerst: Finally fixed it to green, but had to fixate simplecov to an old version.

n-rodriguez (approved):

n-rodriguez:
@codebycliff I think this one can be merged then https://github.com/drapergem/draper/pull/897

Then we'll have a green CI ready to be tested against Ruby 3.0 \o/

n-rodriguez:
Also it could be great to have appraisal to test against different versions of Rails. wdyt?

codebycliff:
I think that would be awesome if someone wants to whip it up. We should also test against Ruby 3, but didn't want to make that change as part of the CI change.

codebycliff (approved):
Thank you for this! Sorry for my late reply on these.

n-rodriguez:
> I think that would be awesome if someone wants to whip it up. We should also test against Ruby 3, but didn't want to make that change as part of the CI change.

I can do a PR for that.

PR #835, opened 2018-09-02
use current module's namespace for decorates_association

john-denisov:
## Description
I have the same issue as author of the issue #809: Our app has two set of decorators: top level decorators and `Admin::*` decorators. Right now I have to explicitly specify `:with` option for every `decorates_association` call inside my admin decorators. This PR helps to simplify the process. All associations will be decorated using decorators from the same namespace as current decorator. 
#### Bonus: 
It is now possible to decorate object with namespaced decorator using `object.decorate(namespace: "Admin"` method instead of using `Admin::ObjectDecorator.new(object)` approach which is harder to automate.

I don't like my current implementation too much. All that passing of `:namespace` option from `DecoratedAssociation#initialize` deep into `Decoratable.decorator_class`looks too complicated . It would be easier to build namespaced decorator class name inside `DecoratedAssociation#initialize` and set it as value of `:with` option (if not present). This approach would look a lot cleaner but from architectural point of view it seems wrong because it is not `DecoratedAssociation`'s responsibility to build decorator class names.

## Testing
You can use any rails project that has `has_many->belongs_to` relation to test this feature. Let's assume that you have `User` model that `has_many :comments`. Sounds a bit too familiar, eh? 😄 
```ruby
# app/decorators/admin/user_decorator.rb
class Admin::UserDecorator < Draper::Decorator
  decorates_association :comments
end

# app/decorators/admin/comment_decorator.rb
class Admin::CommentDecorator < Draper::Decorator
  decorates_association :user
end

comment = Comment.first.decorate(namespace: "Admin")
decorated_comment.class # => Admin::CommentDecorator

decorated_user = decorated_comment.user
decorated_user.class # => Admin::UserDecorator

decorated_comment = decorated_user.comments.first
decorated_comment.class # => Admin::CommentDecorator
```

## To-Dos
- Check if we need to document this feature in README

## References
[Issue #809](https://github.com/drapergem/draper/issues/809)

Issue #809, opened 2017-06-12
Would be nice to have a :module option to decorates_association


itkin:
Hello,

I use drapper to decorates models in the context of my app and its admin interface. 

Say for a model class Document i have 2 different decorators `DocumentDecorator` and `Admin::DocumentDecorator`. 

On the other hand i heavily use STI. Currently `decorates_association` handle STI pretty well but not when you have more than one decorator per STI class.

Would be nice to have a :module option to decorates_association, eventually settable at the decorator class level for all associations, in order to handle STI associations more nicely : 

```ruby
class UserDecorator < ApplicationDecorator
  decorates_association :documents
end

class DocumentDecorator  < ApplicationDecorator
end

class BillingDocumentDecorator  < DocumentDecorator
end

class KycDocumentDecorator  < DocumentDecorator
end

class Admin::UserDecorator < UserDecorator
  decorates_association :documents, module: "Admin"
end

class Admin::DocumentDecorator  < DocumentDecorator
end

class Admin::BillingDocumentDecorator  < Admin::DocumentDecorator
end

class Admin::KycDocumentDecorator  < Admin::DocumentDecorator
end

> user = user.decorate(with: Admin::UserDecorator)
> user.documents
> => [kyc_document_decorator_instance, billing_document_decorator_instance, ..]

```

I could try to send you a pull request if you find the feature interesting let me know, thx for this nice gem.

codebycliff:
Thanks for the suggestion!

 I can see the use case you are after. The inference this gem is currently doing causes quite a bit of headaches, so I'm hesitant to add more magic. However, I feel like we can do a better job of handling this case. It seems like it should be possible to infer the module without the addition of a `module` option. In your example above, we could potentially infer the `Admin` module because the `UserDecorator` is under the `Admin` namespace.

I'd be happy to entertain a pull request if you are interested in submitting one.

Alexander-Senko:
> In your example above, we could potentially infer the `Admin` module because the `UserDecorator` is under the `Admin` namespace.

:+1: to that.

If one wants to be explicit about decorator used, I'd suggest adding of `with:` optional parameter to ~`.decorates_association`~ `.decorates_assigned`.

Issue #661, opened 2015-03-27
.decorate is the Devil


jcasimir:
This is the first line of code in the README:

```ruby
Article.find(params[:id]).decorate
```

And I hate it. Scroll through the closed issues and you'll find many of them are centered around making this feature work. It has to guesstimate (*infer*) what the matching decorator class is. It has to hook into the underlying ORM. It has to take different actions when there are chained methods or associations involved. *It's trying to do too much*.

I've been writing and teaching Ruby and Rails for ten years now. One lesson I've learned over and over and over is "*explicit* over *implicit*". And this is a perfect example. The library implementation would be much easier and less buggy if the above were replaced with this:

```ruby
ArticleDecorator.new( Article.find(params[:id])
```

Is it cute? No. Is the intention and execution clear? *Absolutely*.

Is there a compromise? I still like this syntax:

```ruby
ArticleDecorator.find(params[:id])
```

It's clear, easy to implement, and easier to use.

## Why You Care

Now that I got a 2.0 out the door and dealt with all the major issues, I want to work on a 3.0. And the major change will be completely removing `.decorate`, all the tests and code to support it, and the hooks into AR/Mongoid. I'm curious about creating a sister repo/gem, maybe `draper-infer`, which extracts all this functionality out. If you're deadset on using it and interesting in trying to squash the ongoing bugs, then you can do that. But I don't think it's worth it.

Along the way we can continue rolling 2.X versions that support `decorate`. If 3.0 comes together then let's try to make a 2.X with deprecation warnings to ease the transition.

In the end I hope to have a library that's (a) easy to use, (b) easier to work on, and (c) more resistant to bugs.

jcasimir:
I should have made some fake issues so this would be #666.

fluxusfrequency:
LMK if you want any help, Jeff.

emaiax:
@jcasimir the new syntax is very sweet and similar to AR, but what is going to happen with `decorate_collection`?

jdickey:
(pumping fist in air) Oh, yes, *thank you!* I agree that "explicit over implicit" is The Way To Go™; getting there on my active projects has kept me gainfully employed for the better part of a year now.

[...]

Issue #373, opened 2012-12-10
Decorated Association namespace


vizjerai:
It would be nice if the decorated association would allow specifying a module when inferring the decorator it would use.

Example: I have an OrderItem that has a polymorphic type and the decorators that I'm using are under OrderItem module.

``` ruby
class OrderItemDecorator < Draper::Decorator
  decorates_association :type, module: :order_item
end
```

``` ruby
class OrderItem::ProductDecorator < Draper::Decorator
end
```

``` ruby
class OrderItem::ServiceDecorator < Draper::Decorator
end
```

PR #899, opened 2021-01-25
Add :namespace option to decorate

n-rodriguez:
This PR is a rebase on `master` branch of @ivan-denysov work. It also contains some fixes to finalize the implementation. See https://github.com/drapergem/draper/pull/835#issuecomment-571708445

ping @codebycliff 

it closes 
* https://github.com/drapergem/draper/pull/835
* https://github.com/drapergem/draper/issues/809

### DIFF
--- a/lib/draper/collection_decorator.rb
+++ b/lib/draper/collection_decorator.rb
@@ -16,6 +16,10 @@ module Draper
     #   to each item's decorator.
     attr_accessor :context
 
+    # @return [String, nil] decorator namespace to be used to instantiate decorator, and passed
+    #   to each item's decorator.
+    attr_accessor :namespace
+
     array_methods = Array.instance_methods - Object.instance_methods
     delegate :==, :as_json, *array_methods, to: :decorated_collection
 
@@ -27,11 +31,14 @@ module Draper
     # @option options [Hash] :context ({})
     #   extra data to be stored in the collection decorator and used in
     #   user-defined methods, and passed to each item's decorator.
+    # @option options [String, nil] :namespace (nil)
+    #   decorator namespace
     def initialize(object, options = {})
-      options.assert_valid_keys(:with, :context)
+      options.assert_valid_keys(:with, :context, :namespace)
       @object = object
       @decorator_class = options[:with]
       @context = options.fetch(:context, {})
+      @namespace = options.delete(:namespace)
     end
 
     class << self
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -15,12 +15,12 @@ module Draper
     # @param [Hash] options
     #   see {Decorator#initialize}
     def decorate(options = {})
-      decorator_class.decorate(self, options)
+      decorator_class(namespace: options.delete(:namespace)).decorate(self, options)
     end
 
     # (see ClassMethods#decorator_class)
-    def decorator_class
-      self.class.decorator_class
+    def decorator_class(namespace: nil)
+      self.class.decorator_class(namespace: namespace)
     end
 
     def decorator_class?
@@ -55,7 +55,7 @@ module Draper
       # @param [Hash] options
       #   see {Decorator.decorate_collection}.
       def decorate(options = {})
-        decorator_class.decorate_collection(all, options.reverse_merge(with: nil))
+        decorator_class(namespace: options[:namespace]).decorate_collection(all, options.reverse_merge(with: nil))
       end
 
       def decorator_class?
@@ -68,9 +68,11 @@ module Draper
       # `Product` maps to `ProductDecorator`).
       #
       # @return [Class] the inferred decorator class.
-      def decorator_class(called_on = self)
+      def decorator_class(called_on = self, namespace: nil)
         prefix = respond_to?(:model_name) ? model_name : name
-        decorator_name = "#{prefix}Decorator"
+        namespace = "#{namespace}::" if namespace.present?
+
+        decorator_name = "#{namespace}#{prefix}Decorator"
         decorator_name_constant = decorator_name.safe_constantize
         return decorator_name_constant unless decorator_name_constant.nil?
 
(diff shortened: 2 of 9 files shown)


dan-jensen:
@n-rodriguez huge respect for the work you've put into this, it looks like you've done a great job.  I was just struggling with decorating a namespaced model for a while and came across this PR.

I wanted to comment because ended up solving my problem a very simple way: by namespacing the decorator the same way as the model.  In my case, `app/decorators/help/article_decorator.rb` defines `class Help::ArticleDecorator < Draper::Decorator`.  Given that this works, I wonder if it's the best solution because 1) it already exists, 2) it relies on convention over configuration, and 3) it plays nice with Rails autoloaders (both classic and zeitwerk both complain if the file namespace doesn't match the class namespace - "expected {FILE} to define constant {CLASS}").  But I realize your code may have benefits I'm not aware of.  What do you think?

I just pushed a [PR with a namespace example](https://github.com/drapergem/draper/pull/901).  (I don't believe it conflicts with this PR in any way.)  Would love to hear your comments on it because you know so much about this area 🙂

n-rodriguez:
> by namespacing the decorator the same way as the model.
> But I realize your code may have benefits I'm not aware of. What do you think?

It is indeed a solution. But you may not want to do that (namespacing the decorator the same way as the model) but instead use more "conceptual" namespaces.

For instance, my Rails app is splited into 3 different websites (1 Rails app with 2 engines) sharing the same database. So there are 3 decorators namespaces : one for each websites and those namespaces are not bound to a specific model.

It looks like :

```ruby
module BackOffice
  # resides in Rails app : rails_root/app/decorators/back_office/application_decorator.rb
  class ApplicationDecorator < Draper::Decorator
  end

  class UserDecorator < ApplicationDecorator
  end
end

module FrenchSite
  # resides in engine : engine/french_site/app/decorators/french_site/application_decorator.rb
  class ApplicationDecorator < Draper::Decorator
  end

  class FooDecorator < ApplicationDecorator
  end
end

module SpanishSite
  # resides in engine : engine/spanish_site/app/decorators/spanish_site/application_decorator.rb
  class ApplicationDecorator < Draper::Decorator
  end

  class BarDecorator < ApplicationDecorator
  end
end
```

You can only do that with this PR.

n-rodriguez:
Ping @codebycliff

n-rodriguez:
Hi there! Any news?

n-rodriguez:
Ping @codebycliff. It would be nice to have some news...

I hope I've not worked for nothing....

n-rodriguez:
ping @codebycliff

n-rodriguez:
ping @codebycliff

n-rodriguez:
is it possible to have some fucking news?

codebycliff:
@n-rodriguez While I understand your frustration and apologize for the delay, I don't appreciate the language and don't really think anyone in the industry would either. We are all volunteers here – where we are in fact working for nothing. The reason for my delay is two fold:

The first being I simply don't have the time to maintain this gem anymore. My work and personal responsibilities have changed significantly since I took on the maintenance of this gem and I don't have the same amount of free time that I once did. 

The second reason is the nature of this change. This pull request is adding more functionality around inference, which is the largest headache of this gem. When I originally took over maintenance of this gem, the original author (@jcasimir) really wanted me to remove the inference entirely per [this issue](https://github.com/drapergem/draper/issues/661). While I never was able to lead up that effort, I'm hesitant to introduce more functionality around the `#decorate` method. I alluded to this hesitancy in the original issue that was opened. The direction the original author wanted to go was adding functionality like this through a contrib package that adds the functionality. I'd recommend this approach if you need to get this functionality out into the wild quickly. Otherwise, the little time I have to dedicate to this gem will be focused on bug fixes and releasing new versions. 

If you (or anyone else) would like to volunteer to helping maintain the repository going forward, that is something we can discuss. I just don't want to merge new features and not be around to support them. Let me know your thoughts. Again, I apologize for the delay.

tovodeverett:
From the sidelines, a few observations:
* I totally understand the frustration @n-rodriguez expressed (3.5 months is a long time), and while I agree with @codebycliff that the language used in the comment was inappropriate, I can also understand where it comes from.
* I totally understand how life can take over and make it hard to find time for even short efforts.  I don't know what's going on in @codebycliff's life, but things changed dramatically for me when I started having kids.  If I can't work on something during paid time, it's a struggle to decide between scratching a personal itch and spending time with my kids so that they can do things we love to do (skiing, mountain biking, etc.).
* Hopefully @codebycliff's measured response will lead to a de-escalation and attract some new blood into helming the project so that @codebycliff can share the reins with someone who has the time to shepherd the project.  That would be an awesome outcome for those of us who rely upon `draper`, but definitely don't have the time or skills to manage it.

n-rodriguez:
blabla

n-rodriguez:
FYI I've made a fork that implements this feature : https://github.com/jbox-web/draper

Alexander-Senko (commented):
I'm sorry, but I have contradictory thoughts on the implementation. I like the idea, but I'm not quite sure yet on how it should be implemented 🤔

I'm not so strict on _magic_ as @jcasimir is (to say the truth I do love Ruby for its _magic_), but I should agree with @codebycliff that we should be very cautious about the core API not to put it into a mess.

### API

The first thing that puts me on guard here is using _namespaces_ together with explicitly defined decorator classes. It looks inconsistent for me and can lead to a lot of issues due to misunderstanding (as well as bugs, for sure).

### Implementation

Unfortunately, I can't comment the code as far as the original repository is gone. So, I have to put it inline here.

> ```ruby
>       name = name.split('::').last unless name.nil?
> ```

Looks like it breaks support for namespaced models, doesn't it? We'd better had a class lookup here.

### Conclusion

I'm open to further discussions and will go on thinking of the implementation. I'm new to the project and I need to study its cores better not to break a thing or bring inconsistency in.

Alexander-Senko:
A longer history of the subject: #373 → #480 → #494 → #499.

PR #904, opened 2021-03-16
Localize method updated to support new changes

vishalzambre:
Localize method updated to support new changes
https://api.rubyonrails.org/classes/AbstractController/Translation.html#method-i-localize


codebycliff (approved):
Thank you for this.

codebycliff:
I was in a rush to get this merged so I could get a release out and didn't catch that this was a breaking change. I'm going to have to revert this so I can get a release out without breaking changes. Feel free to open this back up and we can continue discussion from there.

PR #912, opened 2021-12-14
Fix broken CI

y-yagi:
## Description

Latest mongoid only supports Ruby >= 2.5
Ref: https://github.com/mongodb/mongoid/pull/5058


codebycliff (approved):
Thank you ❤️

PR #915, opened 2022-03-01
Use https for GitHub link

mishina2228:
## Description
This reduces unnecessary redirection from http://... to https://... when someone visits this gem's homepage via https://rubygems.org/gems/draper or some tools or APIs that use the gem's metadata.


y-yagi (approved):

y-yagi:
@mishina2228 Thanks!

PR #922, opened 2023-02-24
Update README with config to disable generation of decorator files

chain792:
I added information about generator settings to the README because it was missing.


y-yagi (approved):

y-yagi:
@chain792 Thanks!

PR #927, opened 2023-08-01
fix issue with using draper outside of controller/view context

### LINKED ISSUE #926: issue with using draper outside of controller/view context within a rake task or (active)job
If you use (by mistake? or on purpose?) draper outside of a view context (controller/view) then use some route helpers, this seems to work - indeed it does work wherever you test it, except for a few weird places - like in a Rake task or in a Sidekiq/Delayed Job runner. 

But when you test those using an automated test or from `rails c` / `rails r`, they will work.

I've narrowed this down to Draper::ViewContext::BuildStrategy::Full#controller which (if called outside of a controller/view context) will:

1) infer a controller, and
https://github.com/drapergem/draper/blob/89ec4b70f443bd6db526eddf73bddac52d6ca75b/lib/draper/view_context/build_strategy.rb#L39

2) manufacture a request (using a test request) but only if `ActionController::TestRequest` is defined.
https://github.com/drapergem/draper/blob/89ec4b70f443bd6db526eddf73bddac52d6ca75b/lib/draper/view_context/build_strategy.rb#L41

I appreciate, using decorate outside of view context is unusual, but not beyond the bounds of possibility. 

It turns out `ActionController::TestRequest` is possibly deprecated (maybe this is why it's not autoloaded in some cases?)
https://github.com/rails/rails/blob/main/actionpack/lib/action_controller/test_case.rb#L32-L34

But in any case could one just replace it with `ActionDispatch::TestRequest` 

I've done this and it seems to pass the test suite (except for the tests that specifically check the ActionController::TestRequest class). However I think it's not worth supporting EOLd rails in it. I'll send through a PR and can discuss there if there's any interest in merging this.

--- 

examples:

assume you have generated SomeModel, e.g.

`rails g scaffold user email:string`

and you have a rake task such as: 

```ruby
# lib/test_decoration.rake
task :test_request1 => :environment do
  p(defined:  defined?(ActionController::TestRequest))
end

task :test_request2 => :environment do
  p(defined:  defined?(ActionDispatch::TestRequest))
end

task :test_decorator => :environment do
  User.new.decorate.h.users_path
end
```

I've chosen `development` here, but works the same with `test`, `production`, `staging` (with typical rails defaults) 

```
RAILS_ENV=development DISABLE_SPRING=1 rails r "p(defined: defined?(ActionController::TestRequest))"
{:defined=>"constant"}

RAILS_ENV=development DISABLE_SPRING=1 rails r "p(users_path: User.new.decorate.h.users_path)"
{:users_path=>"/users"}

RAILS_ENV=development DISABLE_SPRING=1 rake test_request1
{:defined=>nil}

#but: 
RAILS_ENV=development DISABLE_SPRING=1 rake test_request2
{:defined=>"constant"}

RAILS_ENV=development DISABLE_SPRING=1 rake test_decorator
NoMethodError: undefined method `host' for nil:NilClass
/path/gems/actionpack-6.1.7.4/lib/action_controller/metal/url_for.rb:32:in `url_options'
/path/gems/actionview-6.1.7.4/lib/action_view/routing_url_for.rb:123:in `url_options'
/path/gems/actionpack-6.1.7.4/lib/action_dispatch/routing/route_set.rb:262:in `call'
/path/gems/actionpack-6.1.7.4/lib/action_dispatch/routing/route_set.rb:213:in `call'
/path/gems/actionpack-6.1.7.4/lib/action_dispatch/routing/route_set.rb:326:in `block in define_url_helper'
/path/gems/draper-4.0.2/lib/draper/helper_proxy.rb:32:in `block in define_proxy'
/path/gems/draper-4.0.2/lib/draper/helper_proxy.rb:13:in `method_missing'
/app/lib/tasks/test_decoration.rake:8:in `block in <main>'
/app/Rakefile:12:in `block in execute_with_benchmark'
/app/Rakefile:12:in `execute_with_benchmark'
```

A similar situation will happen if you execute a ActiveJob such as:

```
# app/jobs/some_job.rb
class SomeJob < ApplicationJob
  queue_as :default

  def perform
    p(defined:  defined?(ActionController::TestRequest))
    User.new.decorate.h.users_path
  end
end
```

`SomeJob.perform_later` (raises error in sidekiq/background job runner)

timdiggins:
## Description

Replace `ActionController::TestRequest` with `ActionDispatch::TestRequest`

because
1) soft deprecated
2) will cause very hard to predict/test failures with ActiveJob/Sidekiq/Rake tasks. see #926 

The current PR also drops support for rails < 6.0 
I think this is a good idea (especially as the tests that were dropped were written from a Rails 5.0 perspective, so are very outdated, and brittle) but open to discussion around this if there's any engagement on this PR.

## Testing

See #926 

## To-Dos

- [ ] tests: I have not done additional tests to prove the need for the PR as they are hard to implement and harder to maintain and wouldn't catch regressions in any case.
- [ ] documentation: unnecessary I think

## References
* [GitHub Issue 926](https://github.com/drapergem/draper/issues926)

fixes #926

### DIFF
--- a/lib/draper/view_context/build_strategy.rb
+++ b/lib/draper/view_context/build_strategy.rb
@@ -38,16 +38,8 @@ module Draper
         def controller
           Draper::ViewContext.controller ||= Draper.default_controller.new
           Draper::ViewContext.controller.tap do |controller|
-            controller.request ||= new_test_request controller if defined?(ActionController::TestRequest)
+            controller.request ||= ActionDispatch::TestRequest.create
           end
-        end
-
-        def new_test_request(controller)
-          is_above_rails_5_1 ? ActionController::TestRequest.create(controller) : ActionController::TestRequest.create
-        end
-
-        def is_above_rails_5_1
-          ActionController::TestRequest.method(:create).parameters.first == [:req, :controller_class]
         end
       end
     end
--- a/spec/draper/view_context/build_strategy_spec.rb
+++ b/spec/draper/view_context/build_strategy_spec.rb
@@ -37,29 +37,12 @@ module Draper
 
         expect(controller.request).to be_nil
         strategy.call
-        expect(controller.request).to be_an ActionController::TestRequest
+        expect(controller.request).to be_an ActionDispatch::TestRequest
         expect(controller.params).to be_empty
 
         # sanity checks
         expect(controller.view_context.request).to be controller.request
         expect(controller.view_context.params).to be controller.params
-      end
-
-      it "compatible with rails 5.1 change on ActionController::TestRequest.create method" do
-        ActionController::TestRequest.class_eval do
-          if ActionController::TestRequest.method(:create).parameters.first == []
-            def create controller_class
-              create
-            end
-          end
-        end
-        controller = Class.new(ActionController::Base).new
-        allow(ViewContext).to receive_messages controller: controller
-        strategy = ViewContext::BuildStrategy::Full.new
-
-        expect(controller.request).to be_nil
-        strategy.call
-        expect(controller.request).to be_an ActionController::TestRequest
       end
 
       it "adds methods to the view context from the constructor block" do


tovodeverett:
I have no authority at all with Draper, so I'm sorry that I can't help with getting your PR approved.

I do have one comment - I found the first line of the Description to be rather confusing.

You wrote: "Replace ActionController::TestRequest with ActionController::TestRequest"
I think you may have meant to write, "Replace ActionController::TestRequest with ActionDispatch::TestRequest"

mdudzinski:
I can confirm this PR fixed the problem (that popped up after upgrading to rails 6.1) in my project. 

In my case it was:

```
NoMethodError: undefined method `parameters' for nil:NilClass

@_params ||= Parameters.new(request.parameters)

~/.gem/ruby/3.1.4/gems/actionpack-6.1.7.7/lib/action_controller/metal/strong_parameters.rb:1187:in `params'
```

lib/draper/view_context/build_strategy.rb
```diff
@@ -38,16 +38,12 @@ def call
         def controller
           Draper::ViewContext.controller ||= Draper.default_controller.new
           Draper::ViewContext.controller.tap do |controller|
-            controller.request ||= new_test_request controller if defined?(ActionController::TestRequest)
+            controller.request ||= new_test_request controller
           end
         end
 
         def new_test_request(controller)
-          is_above_rails_5_1 ? ActionController::TestRequest.create(controller) : ActionController::TestRequest.create
```
Alexander-Senko, line 46: Could we just add a line before as a fix for now?

```ruby
          return ActionDispatch::TestRequest.create if defined? ActionDispatch::TestRequest
```

Alexander-Senko (changes requested):
:+1: for dropping support of Rails 5, but could we split it from the fix?

The reason is that support for older Rails & Ruby versions is far more complicated than that of the code changed in this PR. E.g. we still have a CI for Ruby 2.4 that's incompatible with Rails 6.

lib/draper/view_context/build_strategy.rb
Alexander-Senko: Perfect! Thank you for diving deeper into this 🙏

lib/draper/view_context/build_strategy.rb
```diff
@@ -38,16 +38,12 @@ def call
         def controller
           Draper::ViewContext.controller ||= Draper.default_controller.new
           Draper::ViewContext.controller.tap do |controller|
-            controller.request ||= new_test_request controller if defined?(ActionController::TestRequest)
+            controller.request ||= new_test_request controller
```
Alexander-Senko, line 41: Just a small thing: could we put it inline to make things even simpler?

```suggestion
            controller.request ||= ActionDispatch::TestRequest.create
```

Alexander-Senko (changes requested):

Alexander-Senko (approved):
Thanks a lot for your contribution 🙏

PR #817, opened 2017-08-01
Feature: ActiveJob Integration and Documentation

codebycliff:
## Description
This pull request makes it so that decorators implement `GlobalID` so they can be passed to background jobs. While you can now pass a decorated object to a background job, it won't be decorated when it is deserialized. This may lead to some confusion, so it is documented in the README.

## Testing

Try to pass a decorated object to a background job and it should no longer throw `ActiveJob::SerializationError`.

## To-Dos
- [x] tests
- [x] documentation

## References
* [GitHub Issue 663](https://github.com/drapergem/draper/issues/663)
* [GitHub Pull Request #669 containing initial documentation](https://github.com/drapergem/draper/pull/669)

PR #928, opened 2024-02-06
Improved integration with Global ID

Alexander-Senko:
## Description

Implementing `Decorator.find` improves compatibility with Global ID and allows one to use decorated objects in background jobs seamlessly.

## Testing

1. Decorate an object.
2. Serialize the decorated object with Global ID or pass it to a job.
3. Deserialize the object.
4. Check if it's the same one and is still decorated.

## To-Dos
- [x] tests
- [x] documentation

## References
* Resolves [#663](https://github.com/drapergem/draper/issues/663)
* Improves [#817](https://github.com/drapergem/draper/pull/817)

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,9 +672,8 @@ you can include this module manually.
 [Active Job](http://edgeguides.rubyonrails.org/active_job_basics.html) allows you to pass ActiveRecord
 objects to background tasks directly and performs the necessary serialization and deserialization. In
 order to do this, arguments to a background job must implement [Global ID](https://github.com/rails/globalid).
-Decorated objects implement Global ID by delegating to the object they are decorating. This means
-you can pass decorated objects to background jobs, however, the object won't be decorated when it is
-deserialized.
+Decorators implement Global ID.
+This means you can pass decorated objects to background jobs, and get them just as decorated when deserialized.
 
 ## Contributors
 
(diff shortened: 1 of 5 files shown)


y-yagi (changes requested):
Could you rebase with the `master` and run CI?
I tried to run the test on my local with this PR, but some tests failed.

y-yagi:
@Alexander-Senko 
This patch itself looks good to me. But, why did you add  `4.0.3` tag to this PR?  This PR brings a little breaking change. I think it's not a good idea to introduce a breaking change in a patch version update.

Alexander-Senko:
> This patch itself looks good to me.

Thus, could you approve it, please?

> But, why did you add `4.0.3` tag to this PR?

I don't have a strict opinion on that. You can reattach it to a milestone you like. It does nothing for now except organizing the issues until we get access to the gem.

> This PR brings a little breaking change. I think it's not a good idea to introduce a breaking change in a patch version update.

I see it a bit differently — as a know bug, but not an official API 🤔 Patch releases may break hacks and workarounds if they don't follow official API, may not they?

y-yagi:
> I don't have a strict opinion on that. You can reattach it to whatever milestone you like. It does nothing for now except organizing the issues until we get access to the gem.
> I've just started organizing the issues and introduced new milestones. Any help on that is appreciated.

Thanks for your detailed explanation. I understand what the purpose.

> I see it a bit differently — as a know bug, but not an official API 🤔 Patch releases may break hacks and workarounds if they don't follow official API, may not they?

I don't think so. This is documented behavior, you have changed the document, right?
We shouldn't change the documented behavior in changing patch versions and minor versions(as we follows  the semantic versioning strictly).

y-yagi (approved):
LGTM!

Issue #910, opened 2021-06-18
Usage with `Turbo::Broadcastable`


Alexey1100:
Looking for a good practice when using decorated objects inside templates broadcasted with [Turbo::Broadcastable](https://github.com/hotwired/turbo-rails/blob/d89e1a0f36373180806ceab6949fb25e0d74264a/app/models/concerns/turbo/broadcastable.rb#L36).

Since the rendering happens in `Turbo::Streams::ActionBroadcastJob`, the objects are getting serialised (#817) and the decoration is lost.

```ruby
class Clearance < ApplicationRecord
  belongs_to :petitioner, class_name: "Contact"
  belongs_to :examiner,   class_name: "User"

  after_create_commit :broadcast_later

  private
  
    # Sends <turbo-stream action="replace" target="clearance_5"><template><div id="clearance_5">Other partial</div></template></turbo-stream>
    # to the stream named "identity:2:clearances"
    def broadcast_later
      broadcast_replace_later_to examiner.identity, :clearances, target: self, partial: "clearances/other_partial", locals: { clearance: self }
    end
end
```

Alexey1100:
This is the quick solution I'm using currently, overriding the `Turbo::Streams::ActionBroadcastJob` job, as pointed by dhh in https://github.com/hotwired/turbo-rails/issues/190

```ruby
class Turbo::Streams::ActionBroadcastJob < ApplicationJob
  def perform(stream, action:, target:, **rendering)
    rendering[:locals]&.transform_values! do |v|
      v.decorate
    rescue Draper::UninferrableDecoratorError, NoMethodError
      v
    end

    Turbo::StreamsChannel.broadcast_action_to stream, action: action, target: target, **rendering
  end
end
```

PR #929, opened 2024-02-06
Integration with `Turbo::Broadcastable`

Alexander-Senko:
## Description
Overriding defaults for Turbo broadcast jobs allows one to get decorated objects in model partials by default.

## Testing

1. Broadcast an object with Turbo.
2. Check if the local variable passed into the partial is decorated.

## To-Dos
- [x] tests
- [x] documentation

## References
* Resolves [#910](https://github.com/drapergem/draper/issues/910)
* Depends on [#928](https://github.com/drapergem/draper/pull/928)

### DIFF
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -1,4 +1,5 @@
 require 'draper/decoratable/equality'
+require 'draper/compatibility/broadcastable'
 
 module Draper
   # Provides shortcuts to decorate objects directly, so you can do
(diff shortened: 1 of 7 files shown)


lib/draper/decoratable.rb
```diff
@@ -11,6 +12,10 @@ module Decoratable
     extend ActiveSupport::Concern
     include Draper::Decoratable::Equality
 
+    included do
+      prepend Draper::Compatibility::Broadcastable if defined? Turbo
```
y-yagi, line 16: I think `Turbo` is too generic.  For example,  [turbo-mount](https://github.com/skryukov/turbo-mount) gem uses the same namespace and it is possible that an application uses it. Can we use a more unique name of turbo-rails?

y-yagi (approved):
LGTM!

Issue #919, opened 2022-05-17
Scaffold generator: Invalid path when using --model-name=Record


james-em:
Hi,

I'm working on a project where Controllers and routes are namespaced but models are not.

This is how my scaffold are generated:

```
bundle exec rails g model Client first_name:string last_name:string
bundle exec rails g scaffold_controller Admin::Client first_name:string last_name:string --model-name=Client
```

When specifying --model-name=Client, it would be great if the decorator wasn't namespaced and also not in a namespace folder, just like the model.

An easy fix would be editing this file: https://github.com/drapergem/draper/blob/master/lib/generators/rails/decorator_generator.rb
and changing 

```
def create_decorator_file
  template 'decorator.rb', File.join('app/decorators', class_path, "#{file_name}_decorator.rb")
end
```

to

```
def create_decorator_file
  if parent_options[:model_name].present?
    template 'decorator.rb', File.join('app/decorators', "#{file_name}_decorator.rb")
  else
    template 'decorator.rb', File.join('app/decorators', class_path, "#{file_name}_decorator.rb")
  end
end
```

and changing

https://github.com/drapergem/draper/blob/master/lib/generators/rails/templates/decorator.rb

```
  <%- if parent_class_name.present? -%>
class <%= class_name %>Decorator < <%= parent_class_name %>
  <%- else -%>
class <%= class_name %>
  <%- end -%>
```

To not include ModuleName::ModelNameDecorator: class_name.split("::").last

jamesst20:
@Alexander-Senko @y-yagi 

I have taken a look at https://github.com/drapergem/draper/pull/931

I'm with ghost and Alexander on this, I generate my scaffolds exactly the same was as op

```
   bundle exec rails generate model MyModel name:string description:text price:decimal
   bundle exec rails generate scaffold_controller Something::MyModel name:string description:text price:decimal --model-name=MyModel
```

From my understanding, the `decorate` method should infer the decorator based on the model class itself unless `with:` is passed.

Maybe the right approach would be to scaffold the decorator with the model instead of the controller so there wouldn't be a need to check for `--model-name` and if the model scaffold uses a namespace, then it should use it

PR #931, opened 2024-08-31
Generators: allow decorators not to match controller names

Alexander-Senko:
## Description

Decorator will respect `--model-name` passed to `scaffold_controller` or `controller` generator.

That could be of a particular use to solve namespace issues (e.g. when controllers are namespaced and models are not).

The specs have been refactored to reduce and DRY the code.

## Testing

1. Use `controller` or `scaffold_controller` generator with `--model-name` option.
2. Check if the generated decorator matches the model name.

## To-Dos

- [x] tests

## References

* #919

### DIFF
--- a/lib/generators/controller_override.rb
+++ b/lib/generators/controller_override.rb
@@ -1,17 +1,24 @@
 require "rails/generators"
 require "rails/generators/rails/controller/controller_generator"
 require "rails/generators/rails/scaffold_controller/scaffold_controller_generator"
+require "rails/generators/model_helpers"
 
 module Rails
   module Generators
     class ControllerGenerator
+      include Rails::Generators::ModelHelpers
+
+      class_option :model_name, type: :string, desc: "ModelName to be used"
+
       hook_for :decorator, type: :boolean, default: true do |generator|
-        invoke generator, [name.singularize]
+        invoke generator, [options[:model_name] || name]
       end
     end
 
     class ScaffoldControllerGenerator
-      hook_for :decorator, type: :boolean, default: true
+      hook_for :decorator, type: :boolean, default: true do |generator|
+        invoke generator, [options[:model_name] || name]
+      end
     end
   end
 end
--- a/spec/generators/controller/controller_generator_spec.rb
+++ b/spec/generators/controller/controller_generator_spec.rb
@@ -5,19 +5,65 @@ require 'generators/controller_override'
 require 'generators/rails/decorator_generator'
 SimpleCov.command_name 'test:generator'
 
-describe Rails::Generators::ControllerGenerator do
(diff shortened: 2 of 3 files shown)


lib/generators/controller_override.rb
```diff
@@ -1,17 +1,24 @@
...
 require "rails/generators/rails/controller/controller_generator"
 require "rails/generators/rails/scaffold_controller/scaffold_controller_generator"
+require "rails/generators/model_helpers"
 
 module Rails
   module Generators
     class ControllerGenerator
+      include Rails::Generators::ModelHelpers
+
+      class_option :model_name, type: :string, desc: "ModelName to be used"
```
y-yagi, line 11: Hmm. Basically, `controller` doesn't need to know about `model`. And, the current implementation depends on `controller` name for naming(That's why this file overrides `ControllerGenerator` I think). So overring `controller` name by `model_name` looks strange to me. 
How about using `decorator_name`?

Alexander-Senko: `scaffold_controller` already has that interface and I'd prefer `controller` to be uniform with it. Rails already has `--model-name` option and I'd rather use it.

I agree that model dependency for `controller` is dubious, but we already have a `decorator` hook for `controller` that needs to know about a model. That's why `controller` may also need to.

Another option is to remove `decorator` hook from `controller` and add it to `model`. In my opinion, decorators are bound to models rather than controllers or views. However, I'd keep the existing API for now rather than introducing another braking change.

y-yagi: I agree that it might be good to hook my models, not controllers. But, it seems that the current behavior is intended #553.
So generating decorators by controller names is the intended behavior(At least, so far). Setting a decorator name by `--model-name` is inconsistent with that behavior I think.

Alexander-Senko: > So generating decorators by controller names is the intended behavior.

Yes, it is. And it's not being changed.

Models are also generated based on controller names, but we have `--model-name` to override that when needed.

> `scaffold_controller` already has that interface…

So, why shouldn't a decorator respect it as well when it's specified explicitly?

I'm afraid, I'm missing your point… What is your solution for #919?

y-yagi: > Yes, it is. And it's not being changed.

I think this PR changes the behavior. For example, if you run `./bin/rails g scaffold admin/book name:string --model-name book`, the result will be change between `master` branch and this PR.

To be honest, I don't think #919 is a bug. That is the intended behavior of the current generator structure(I can't understand why `draper` only hooks for `controller` though).
But of course, I understand the user's expectations. So adding `decorator_name`(or something like that) to override decorators name is a good solution I think.

Issue #827, opened 2017-12-11
model.assocations.decorate loads the association even if it is preloaded (AR)


PragTob:
Hi there,

first and foremost - thanks for draper :tada: 

In our application I run into the problem that I want to explicitly decorate an association not as part of the decorator definition (sadly this easily leads to performance problems) so we want to do `model.association.decorate` but this seems to reload the objects from the database even if they were preloaded before.

See this snippet:

```
[13] pry(main)> ts = TourSchedule.includes(:tour_schedule_plans).find(104281); nil
  TourSchedule Load (1.0ms)  SELECT  "tour_schedules".* FROM "tour_schedules" WHERE "tour_schedules"."id" = $1 LIMIT $2  [["id", ID], ["LIMIT", 1]]
  TourSchedulePlan Load (1.3ms)  SELECT "tour_schedule_plans".* FROM "tour_schedule_plans" WHERE "tour_schedule_plans"."tour_schedule_id" = ID
=> nil
[14] pry(main)> ts.tour_schedule_plans.loaded?
=> true
[15] pry(main)> ts.tour_schedule_plans.to_a; nil
=> nil
[16] pry(main)> ts.tour_schedule_plans.decorate.to_a; nil
  TourSchedulePlan Load (0.8ms)  SELECT "tour_schedule_plans".* FROM "tour_schedule_plans" WHERE "tour_schedule_plans"."tour_schedule_id" = $1  [["tour_schedule_id", ID]]
=> nil
```

Draper version in use is 3,0.0.

Somewhat related issues (associations and preloading) are:
* #646 (the way we decorate just differs)
* #812 

edit: FYI mapping over all objects and decorating them instead seems to be a fine workaround as indicated by early testing `ts.tour_schedule_plans.map &:decorate`

Thanks!

brunohkbx:
@codebycliff I don't think there is any way to avoid this query when `.decorate` calls `.all` even if it was preloaded :thinking: 
https://github.com/drapergem/draper/blob/master/lib/draper/decoratable.rb#L58

Alexander-Senko:
Duplicates #646. The reason is the same as the one for #706.

Alexander-Senko:
@brunohkbx, thanks for the point :+1:

Issue #706, opened 2015-11-10
.decorate on an ActiveRecord::Associations::CollectionProxy ignores current value


riffraff:
This is probably related to https://github.com/drapergem/draper/issues/646, but my specific problem is with unpersisted objects, i.e. if there is an unsaved object in an association in the same example code (`user has_many posts`), we get

```
>> user.posts.build title: 'unsaved post'
... 
>> user.posts.map(&:name))
=> ["saved post", "unsaved post"]
>> user.posts.decorate.map(&:name))
=> ["saved post"]
```

This proves problematic when working with decorated values in nested forms, where if there is a validation error the form shows data as-is-in-db rather than as-is-in-this-page.

While there is a workaround (doing a `.map(&:decorate)` step explicitly), it seems like this is surprising behaviour, is it by design or accidental?

Mohit-Pawar:
@riffraff 
 By design we should use Decorate Associated Objects and Collection Decorator

so if user has_many posts
we should create UserDecorator and PostsDecorator as follows.

class UserDecorator < Draper::Decorator
  delegate_all
  decorates_association :post, with: PostsDecorator
end

class PostsDecorator < Draper::CollectionDecorator
  delegate_all
end

so when you try
>> user.posts.build title: 'unsaved post'

>> user.posts.map(&:name))
=> ["saved post", "unsaved post"] => correct ouput no need to use .map(&decorate)

Issue #646, opened 2014-12-19
Draper forces DB queries on decorated model's associations


swiercz:
Assume:
User has_many posts
Post has_many comments

When fetching users from DB with preloading of posts and comments:

``` ruby
@users = User.where(....).includes(posts: [:comments]).decorate
```

The AR properly preloads users, posts and comments (3 queries) but decorated users' collection forces new DB queries when rendering view.

It seems that objects from decorated collection ignore the fact that associations are preloaded by AR and fore new DB query for each one. This results with a performance drop.

RABL:

``` rabl
child @users do
  extends 'users/_base'
end
```

User partial

``` rabl
attributes :email
child :posts do
  extends 'posts/_base'
end
```

Post partial

``` rabl
attributes :title
child :posts do
  extends 'comments/_base'
end
```

Comment partial

``` rabl
attributes :text, :created_at
```

User decorator

``` ruby
class UserDecorator < Draper::Decorator
  delegate_all
  def posts
    object.posts.decorate
  end
end
```

Post decorator

``` ruby
class PostDecorator < Draper::Decorator
  delegate_all
  def comments
    object.comments.decorate
  end
end
```

dimroc:
:+1: 

We're pretty aggressive with our eager loading so this is causing a pretty big hit to perf on certain views.

jtomaszewski:
Has this been fixed? I'm experiencing the same issue as @dimroc on draper 2.1.0 and I need to use the same workaround, to avoid it:

> Decorating the individual elements rather than the collection avoids this issue.

darrinholst:
I just noticed seeing this as well. @jcasimir sorry for the 2 year old question, but do you remember why this was closed?

seanlinsley:
Given that there isn't a commit or PR linked to this ticket, I suspect it's still an issue.

codebycliff:
I was able to reproduce this behavior and I fixed it by using `decorates_associated` instead of overriding the method:

```ruby
class UserDecorator < Draper::Decorator
  delegate_all
  decorates_associated :posts
end

class PostDecorator < Draper::Decorator
  delegate_all
  decorates_associated :comments
end
```

Were the rest of you experiencing this issue also overriding the association method in the decorator? If so, you should be able to change it to the code above. If there is another scenario that is causing this, please let me. If you could provide an example app, that would be even better.

dimroc:
I tried this in two projects with version draper 3.0.0.pre1.

[...]

PR #932, opened 2024-09-04
Associations: `#decorate` queries DB every time

Alexander-Senko:
`ActiveRecord::Associations::CollectionProxy#decorate` ignores `target` (be it already loaded or not) and loads the association again every time it's called.

That's why unsaved records get missing from the decorated collection.

## Testing

* `it 'returns a decorated collection'`
  1. Ensure a record and some records associated with it are persisted.
  2. Get the decorated association: `record.associated_records.decorate`.
  3. Ensure that it matches `record.associated_records`.

* `it 'uses cached records'`
  1. Ensure a record and some records associated with it are persisted.
  2. Get the association: `record.associated_records`.
  3. Get the decorated association: `record.associated_records.decorate`.
  4. Ensure the association was loaded only once.

* `it 'caches records'`
  1. Ensure a record and some records associated with it are persisted.
  2. Get the decorated association: `record.associated_records.decorate` (twice).
  3. Get the association: `record.associated_records`.
  4. Ensure the association was loaded only once.

* `it 'respects unsaved records'`
  1. Ensure a record and some records associated with it are persisted.
  2. `build` new associated records.
  3. Get the decorated association: `record.associated_records.decorate`.
  4. Ensure that it matches both `record.associated_records` and the new ones.

## To-Dos

- [ ] tests

## References

* #518
* #538
* #588
* #646
* #706
* #827

### DIFF
--- a/lib/draper/decoratable.rb
+++ b/lib/draper/decoratable.rb
@@ -12,6 +12,8 @@ module Draper
     extend ActiveSupport::Concern
     include Draper::Decoratable::Equality
 
+    autoload :CollectionProxy, 'draper/decoratable/collection_proxy'
+
     included do
       prepend Draper::Compatibility::Broadcastable if defined? Turbo::Broadcastable
     end
--- a/draper.gemspec
+++ b/draper.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'ammeter'
   s.add_development_dependency 'rake'
   s.add_development_dependency 'rspec-rails'
+  s.add_development_dependency 'rspec-activerecord-expectations', '~> 1.2.0'
   s.add_development_dependency 'minitest-rails'
   s.add_development_dependency 'capybara'
   s.add_development_dependency 'active_model_serializers', '>= 0.10'
--- a/lib/draper.rb
+++ b/lib/draper.rb
@@ -51,12 +51,6 @@ module Draper
     end
   end
 
-  def self.setup_orm(base)
-    base.class_eval do
-      include Draper::Decoratable
-    end
-  end
-
   class UninferrableDecoratorError < NameError
     def initialize(klass)
       super("Could not infer a decorator for #{klass}.")
--- a/lib/draper/decoratable/collection_proxy.rb
+++ b/lib/draper/decoratable/collection_proxy.rb
@@ -1,0 +1,15 @@
+module Draper
+  module Decoratable
+    module CollectionProxy
+      # Decorates a collection of objects. Used at the end of a scope chain.
+      #
+      # @example
+      #   company.products.popular.decorate
+      # @param [Hash] options
+      #   see {Decorator.decorate_collection}.
+      def decorate(options = {})
+        decorator_class.decorate_collection(load_target, options.reverse_merge(with: nil))
+      end
+    end
+  end
+end
--- a/lib/draper/railtie.rb
+++ b/lib/draper/railtie.rb
@@ -33,10 +33,14 @@ module Draper
     end
 
     initializer 'draper.setup_orm' do
-      [:active_record, :mongoid].each do |orm|
-        ActiveSupport.on_load orm do
-          Draper.setup_orm self
-        end
+      ActiveSupport.on_load :active_record do
+        include Draper::Decoratable
+
+        ActiveRecord::Associations::CollectionProxy.include Draper::Decoratable::CollectionProxy
+      end
+
+      ActiveSupport.on_load :mongoid do
+        include Draper::Decoratable
       end
     end
 
--- a/spec/dummy/app/decorators/comment_decorator.rb
+++ b/spec/dummy/app/decorators/comment_decorator.rb
@@ -1,0 +1,13 @@
+class CommentDecorator < Draper::Decorator
+  delegate_all
+
+  # Define presentation-specific methods here. Helpers are accessed through
+  # `helpers` (aka `h`). You can override attributes, for example:
+  #
+  #   def created_at
+  #     helpers.content_tag :span, class: 'time' do
+  #       object.created_at.strftime("%a %m/%d/%y")
+  #     end
+  #   end
+
+end
--- a/spec/dummy/app/models/comment.rb
+++ b/spec/dummy/app/models/comment.rb
@@ -1,0 +1,3 @@
+class Comment < ApplicationRecord
+  belongs_to :post
+end
--- a/spec/dummy/app/models/post.rb
+++ b/spec/dummy/app/models/post.rb
@@ -3,5 +3,7 @@ require 'turbo/broadcastable' if defined? Turbo::Broadcastable # HACK: looks wei
 class Post < ApplicationRecord
   # attr_accessible :title, :body
 
+  has_many :comments
+
   broadcasts if defined? Turbo::Broadcastable
 end
--- a/spec/dummy/db/migrate/20240907041839_create_comments.rb
+++ b/spec/dummy/db/migrate/20240907041839_create_comments.rb
@@ -1,0 +1,9 @@
+class CreateComments < ActiveRecord::Migration[6.1]
+  def change
+    create_table :comments do |t|
+      t.references :post, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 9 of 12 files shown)


Alexander-Senko:
I'm sorry, but the test environment looks like a mess for me 😕 I spent some time figuring out how it's supposed to test Active Record integration — and found nothing looking like a valid example. Was it ever tested? 🤔

So, I've tested it manually. Could we pull this in without a test coverage for now?

y-yagi:
> I'm sorry, but the test environment looks like a mess for me 😕 I spent some time figuring out how it's supposed to test Active Record integration — and found nothing looking like a valid example. Was it ever tested? 🤔

I'm sorry I've misunderstood something, but I think we can add these kinds of specs under the `spec/dummy` you already did in  #929.  At least, the following spec works as expected.

```diff
diff --git a/spec/dummy/app/decorators/comment_decorator.rb b/spec/dummy/app/decorators/comment_decorator.rb
new file mode 100644
index 0000000..8257739
--- /dev/null
+++ b/spec/dummy/app/decorators/comment_decorator.rb
@@ -0,0 +1,2 @@
+class CommentDecorator < Draper::Decorator
+end
\ No newline at end of file
diff --git a/spec/dummy/app/models/comment.rb b/spec/dummy/app/models/comment.rb
new file mode 100644
index 0000000..8b86c56
--- /dev/null
+++ b/spec/dummy/app/models/comment.rb
@@ -0,0 +1,3 @@
+class Comment < ApplicationRecord
+  belongs_to :post
+end
diff --git a/spec/dummy/app/models/post.rb b/spec/dummy/app/models/post.rb
index 6352fdb..d143128 100644
--- a/spec/dummy/app/models/post.rb
+++ b/spec/dummy/app/models/post.rb
@@ -4,4 +4,6 @@ class Post < ApplicationRecord
   # attr_accessible :title, :body

   broadcasts if defined? Turbo::Broadcastable
+
+  has_many :comments
 end
diff --git a/spec/dummy/db/schema.rb b/spec/dummy/db/schema.rb
index 9aebf2c..8213494 100644
--- a/spec/dummy/db/schema.rb
+++ b/spec/dummy/db/schema.rb
@@ -18,4 +18,9 @@ ActiveRecord::Schema.define(version: 20121019115657) do
     t.datetime "updated_at", null: false
   end

+  create_table "comments", force: true do |t|
+    t.integer "post_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
 end
diff --git a/spec/dummy/spec/models/post_spec.rb b/spec/dummy/spec/models/post_spec.rb
index 8bc05a0..c8dc10e 100644
--- a/spec/dummy/spec/models/post_spec.rb
+++ b/spec/dummy/spec/models/post_spec.rb
@@ -18,4 +18,13 @@ RSpec.describe Post do
       }
     end
   end if defined? Turbo::Broadcastable
+
+  describe 'association' do
+    it 'respects unsaved records' do
+      post = Post.create!
+      Comment.create!(post: post)
+      post.comments.build
+      expect(post.comments.decorate.count).to eq 2
+    end
+  end
 end
```

lib/draper/decoratable.rb
```diff
@@ -12,8 +12,14 @@ module Decoratable
     extend ActiveSupport::Concern
     include Draper::Decoratable::Equality
 
+    autoload :CollectionProxy, 'draper/decoratable/collection_proxy'
+
     included do
       prepend Draper::Compatibility::Broadcastable if defined? Turbo::Broadcastable
+
+      if defined? ActiveRecord
+        ActiveRecord::Associations::CollectionProxy.include CollectionProxy
```
y-yagi, line 21: This file is loaded when ORM is Mongoid too.  So, I think a loading file only for Active Record isn't good here.
How about loading a file in `railtie.rb` like the following. 

```diff 
diff --git a/lib/draper/compatibility/active_record/collection_proxy.rb b/lib/draper/compatibility/active_record/collection_proxy.rb
new file mode 100644
index 0000000..f1ea1e2
--- /dev/null
+++ b/lib/draper/compatibility/active_record/collection_proxy.rb
@@ -0,0 +1,17 @@
+module Draper
+  module Compatibility
+    module ActiveRecord
+      module CollectionProxy
+        # Decorates a collection of objects. Used at the end of a scope chain.
+        #
+        # @example
+        #   company.products.popular.decorate
+        # @param [Hash] options
+        #   see {Decorator.decorate_collection}.
+        def decorate(options = {})
+          decorator_class.decorate_collection(load_target, options.reverse_merge(with: nil))
+        end
+      end
+    end
+  end
+end
diff --git a/lib/draper/railtie.rb b/lib/draper/railtie.rb
index 556c15a..78f1dfd 100644
--- a/lib/draper/railtie.rb
+++ b/lib/draper/railtie.rb
@@ -36,6 +36,11 @@ module Draper
       [:active_record, :mongoid].each do |orm|
         ActiveSupport.on_load orm do
           Draper.setup_orm self
+
+          if orm == :active_record
+            require 'draper/compatibility/active_record/collection_proxy'
+            ActiveRecord::Associations::CollectionProxy.include Draper::Compatibility::ActiveRecord::CollectionProxy
+          end
         end
       end
     end
```

Alexander-Senko: > This file is loaded when ORM is Mongoid too.

That's why `if defined? ActiveRecord` is here. It's similar to `if defined? Turbo::Broadcastable` a line before.

> So, I think a loading file only for Active Record isn't good here.

I'm sorry, but I don't get the point 🤔 `decoratable` isn't loaded just for AR, but `collection_proxy` is, as it is in your example too.

Logically, it's a part of `setup_orm` too. That's why I didn't put into the initializer directly.

y-yagi: I'm not sure the philosophy of Draper correctly yet, but I think Draper tries to use the same codes for all ORM because there are no `defined?(ActiveRecord)`  and `defined?(Mongoid)`. I think we should follow it and add `defined?(ActiveRecord)` would against it.

So I think `lib/draper/railtie.rb` is a good place to add that because that's the only place to consider the ORM now.

Alexander-Senko: > I think `lib/draper/railtie.rb` is a good place to add that

Why not `setup_orm` then?

y-yagi: `setup_orm` only knows a base class, but this PR needs to extend another class(doesn't touch a base class). So I think `setup_orm` isn't a good place to do it.
Or, adding another setup method like `setup_active_record` or `active_record_integration` maybe better because we already have `setup_action_controller` and `minitest_integration`(`rspec_integration`).

Alexander-Senko: > `setup_orm` only knows a base class, but this PR needs to extend another class(doesn't touch a base class).

So what? This still could be use to specify *dependencies* of that class.

> So I think `setup_orm` isn't a good place to do it.

Let's say it again: `setup_orm` isn't a good place to set up Active Record ORM.

> Or, adding another setup method like `setup_active_record` or `active_record_integration` maybe better because we already have `setup_action_controller` and `minitest_integration`(`rspec_integration`).

We have a lot of other mess and inconsistencies I'd like to refactor. But for now, I'd like to fix all the bugs in a backlog. I'm fine if someone does it in theirs own way. I'm afraid, having discussions on style in every PR is too much for me, sorry.

y-yagi: Hmm. That isn't a minor thing for me. It's essential to keep code clean. But, I understand your opinion. I don't decline this. 

By the way, do you have a plan to add specs to this PR? I would like to keep adding specs for PRs that change behavior.

Alexander-Senko: > It's essential to keep code clean.

It's essential for me either. But it's already far from being clean. So, I try to fix bugs first and do my best in keeping off refactoring yet, minimizing diffs. Otherwise, I'd refactor initializers as well.

> By the way, do you have a plan to add specs to this PR?

Yes, sure. I've already reacted 👍 for that. Nevertheless of will you approve the PR or not, I'll make it complete.

Alexander-Senko: > `Draper::Compatibility::ActiveRecord::CollectionProxy`

It's not a _compatibility_, but our business logic.

y-yagi (approved):
LGTM!

PR #933, opened 2024-09-12
Revert class spoofing

Alexander-Senko:
Decorators shouldn't pretend to be instances of the underlying model  classes for the following reasons:

1. It's terribly wrong to pretend being a class without implementing  it's interface. It can break third party code.
2. Hacking Ruby core methods can lead to unexpected and hard to  understand behavior. It should be avoided without a strong reason  behind.
3. I'd prefer compatibility patches to be narrow-scoped and live in  there own modules.

## References

* #72
* #110
* #417
* #497
* #859

### DIFF
--- a/lib/draper/collection_decorator.rb
+++ b/lib/draper/collection_decorator.rb
@@ -61,12 +61,6 @@ module Draper
 
     alias :decorated_with? :instance_of?
 
-    def kind_of?(klass)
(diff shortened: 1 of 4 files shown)


y-yagi (approved):
I agree that there is no reason to maintain this behavior. This would be a breaking change, but I think it's enough to write a document(like uprading guide) when we needed.

LGTM!

PR #935, opened 2024-09-25
Workflows / CI: fixed & refactored

Alexander-Senko:
## Fixed

* Ruby 2.6 (see hotwired/turbo-rails/issues/681).

## Changed

* Moved Rails versions to CI `matrix`.
* Both edge and outdated Ruby / Rails versions may fail.
* Use the latest RubyGems.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,27 +2,42 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - master
+  pull_request:
 
 permissions:
   contents: read
 
 jobs:
   rspec:
-    runs-on: ubuntu-20.04
+    name: >-
+      rspec (${{ matrix.ruby }})
+
+    runs-on: ubuntu-latest
+
     strategy:
       fail-fast: false
       matrix:
         ruby:
-          - '3.3'
-          - '3.2'
-          - '3.1'
-          - '3.0'
-          - '2.7'
-          - '2.6'
-          - '2.5'
-          - '2.4'
+          - 3.3
+          - 3.2
+          - 3.1
+        rails:
+          - 6.1 # TODO: move to 7.2
+        include:
+          # Edge
+          - { ruby: 'head', rails: 'edge', allow-fail: true }
+          # Outdated
+          - { ruby: '3.0',  rails: '6.1' } # TODO: move to 7.1
+          - { ruby: '2.7',  rails: '6'   } # RSpec AR Expectations support Rails 7.1 since Ruby 3.0
+          - { ruby: '2.6',  rails: '6'   }
+          - { ruby: '2.5',  rails: '6'   }
+          - { ruby: '2.4',  rails: '5'   }
+
+    env:
+      RAILS_VERSION: "${{ matrix.rails }}"
 
     services:
       mongodb:
@@ -31,13 +46,13 @@ jobs:
           - 27017:27017
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          rubygems: latest
           bundler-cache: true
 
       - name: RSpec & publish code coverage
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,27 +2,36 @@
 name: CI
 
 on:
-  - push
```
Alexander-Senko, line 5: Otherwise, it runs every job twice on every _push_ to a pull request.

What did we need that `push` trigger for in the first place?

.github/workflows/ci.yml
```diff
@@ -2,27 +2,36 @@
...
   - pull_request
 
 permissions:
   contents: read
 
 jobs:
   rspec:
-    runs-on: ubuntu-20.04
+    name: ${{ format('rspec ({0})', matrix.ruby) }}
```
Alexander-Senko, line 12: I don't like the name, but we need it to match with the _required_ workflows.

Does anyone have an access to change them? Otherwise, we have to support outdated ones for ages 😕

.github/workflows/ci.yml
```diff
@@ -46,3 +58,9 @@ jobs:
           CC_TEST_REPORTER_ID: b7ba588af2a540fa96c267b3655a2afe31ea29976dc25905a668dd28d5e88915
         with:
           coverageCommand: bin/rake
+        continue-on-error: ${{ matrix.allow-fail }}
+        id: test
+
+      - name: >-
+          Test outcome: ${{ steps.test.outcome }}
+        run: echo NOOP # every step must define a `uses` or `run` key
```
Alexander-Senko, line 66: To see the actual result of tests without looking through logs.

Alexander-Senko (commented):
This combines two things actually:

1. A much needed fix for every PR to pass tests.
2. Refactoring for easier maintenance of the version matrix.

.github/workflows/ci.yml
```diff
@@ -2,27 +2,39 @@
...
   - pull_request
 
 permissions:
   contents: read
 
 jobs:
   rspec:
-    runs-on: ubuntu-20.04
+    name: >-
+      rspec (${{ matrix.ruby }})
```
Alexander-Senko, line 13: I don't like the name, but we need it to match with the required workflows.

Does anyone have an access to change them? Otherwise, we have to support outdated ones for ages 😕

.github/workflows/ci.yml
```diff
@@ -2,27 +2,36 @@
...
-          - '3.0'
-          - '2.7'
-          - '2.6'
-          - '2.5'
-          - '2.4'
+          - 3.3
+          - 3.2
+          - 3.1
+        rails:
+          - 6.1 # TODO: move to 7.2
```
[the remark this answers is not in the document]
Alexander-Senko: #936 handles that.

.github/workflows/ci.yml
y-yagi: I can't see the repository setting of `draper`, but if someone allows to commit to the default branch(`master`) directly, we should have this setting to run CI in that case. I think it's better to limit the branch to `master` instead of remove.

```yml
  push:
    branches:
      - master
```

.github/workflows/ci.yml
```diff
@@ -2,27 +2,39 @@
...
+        include:
+          # Edge
+          - { ruby: 'head', rails: 'edge', allow-fail: true }
+          # Outdated
+          - { ruby: '3.0',  rails: '6.1',  allow-fail: true } # TODO: move to 7.1
+          - { ruby: '2.7',  rails: '6',    allow-fail: true } # RSpec AR Expectations support Rails 7.1 since Ruby 3.0
+          - { ruby: '2.6',  rails: '6',    allow-fail: true }
+          - { ruby: '2.5',  rails: '6',    allow-fail: true }
```
y-yagi, line 33: I do not think it is a good idea to use `continue-on-error` in these cases.

This setting makes it impossible to notice that a test has failed, which would make the CI meaningless.
I think it's better to consider drop old versions support instead of using this(I think it's better to only suport not EOL versions).

y-yagi (approved):
I left some comments, but as is OK. LGTM!

.github/workflows/ci.yml
Alexander-Senko: Though `master` is protected, I'll include this.

.github/workflows/ci.yml
Alexander-Senko: > This setting makes it impossible to notice that a test has failed…

Harder to, but not impossible. One still can check for `Test success` or `Test failure` in the steps.

> I think it's better to consider drop old versions support instead of using this(I think it's better to only suport not EOL versions).

I'd like to indeed, but…

1. It's possible for the next _major_ version only.
2. We still have **no access to the settings** to do that 😕

Alexander-Senko:
> LGTM!

I'm sorry, but another approval is needed due to suggested modifications made.

.github/workflows/ci.yml
Alexander-Senko: > I do not think it is a good idea to use `continue-on-error` in these cases.

I've dropped this for old Rubies until we drop support for them.

y-yagi (approved):

PR #939, opened 2025-01-06
CI against Ruby 3.4

y-yagi:
## Description

Ruby 3.4 has been released. I would like to add Ruby 3.4 to CI to confirm whether `draper` works well with it.

### DIFF
--- a/spec/draper/decorator_spec.rb
+++ b/spec/draper/decorator_spec.rb
@@ -451,7 +451,7 @@ module Draper
       it "includes the context" do
         decorator = Decorator.new(double, context: {foo: "bar"})
 
-        expect(decorator.inspect).to include '@context={:foo=>"bar"}'
+        expect(decorator.inspect).to include "#{@context={:foo=>"bar"}}"
       end
 
       it "includes other instance variables" do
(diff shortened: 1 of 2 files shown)


spec/draper/decorator_spec.rb
```diff
@@ -451,7 +451,7 @@ module Draper
       it "includes the context" do
         decorator = Decorator.new(double, context: {foo: "bar"})
 
-        expect(decorator.inspect).to include '@context={:foo=>"bar"}'
+        expect(decorator.inspect).to include "#{@context={:foo=>"bar"}}"
```
y-yagi, line 454: Ruby 3.4 has changed the format of `Hash.inspect`. So I wrapped this with `#{}` to work both Ruby 3.4 and older.
Ref: https://bugs.ruby-lang.org/issues/20433

Alexander-Senko, line 454:
```suggestion
        expect(decorator.inspect).to include "@context=#{{foo: "bar"}}"
```

Alexander-Senko (changes requested):
I didn't test my suggestion, but hope it to be valid.

spec/draper/decorator_spec.rb
y-yagi: It may work. I will apply.

y-yagi:
It seems that Ruby 2.4 is failing with another issue. I will check.

Alexander-Senko:
> It seems that Ruby 2.4 is failing with another issue. I will check.

Maybe the time has come to let it fail?

Alexander-Senko (approved):

PR #945, opened 2025-02-24
Remove unused files from package

krororo:
## Description

The `.gemspec` file to optimize the gem package size and structure. The changes include:

- Modified `files` to exclude test files, Gemfile, and CI related files from the package.
- Removed the deprecated `test_files` specification as it is no longer recommended.

These changes reduce the extracted package size from from 1.1MB to 340KB.

Benefits:

- Shorter download times for users
- Reduced container sizes when the gem is included

## References

- https://github.com/rubygems/guides/issues/90
- https://github.com/rubygems/bundler/pull/3207
- [Template for the *.gemspec file generated by bundle gem command](https://github.com/rubygems/rubygems/blob/bundler-v2.6.5/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt)

### DIFF
--- a/draper.gemspec
+++ b/draper.gemspec
@@ -10,8 +10,15 @@ Gem::Specification.new do |s|
   s.description = "Draper adds an object-oriented layer of presentation logic to your Rails apps."
   s.license     = "MIT"
 
-  s.files         = `git ls-files`.split("\n")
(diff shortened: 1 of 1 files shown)


y-yagi (approved):
This change makes sense to me. There is no reason to add test files to a gem package.

y-yagi:
I will merge this after CI would stable.

PR #946, opened 2025-03-05
Fix Broken CI

y-yagi:
This updates dependencies for old Ruby versions to fix the build.
Ref: https://github.com/drapergem/draper/actions/runs/13019407035/job/36316404382

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,10 @@ gemspec
 gem 'puma'
 
 platforms :ruby do
-  if RUBY_VERSION >= "2.5.0"
+  if RUBY_VERSION >= "3.0.0"
     gem 'sqlite3'
+  elsif RUBY_VERSION >= "2.5.0"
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -5,8 +5,10 @@ gemspec
 gem 'puma'
 
 platforms :ruby do
-  if RUBY_VERSION >= "2.5.0"
+  if RUBY_VERSION >= "3.0.0"
     gem 'sqlite3'
+  elsif RUBY_VERSION >= "2.5.0"
+    gem 'sqlite3', '~> 1.4.0'
```
y-yagi, line 11: Without this change, sqlite3 fails to install.

```
sqlite3-1.7.3-x86_64-linux requires ruby version < 3.4.dev, >= 3.0, which is incompatible with the current version, ruby
2.7.7p221
```

Alexander-Senko (approved):

Issue #951, opened 2025-11-13
4.0.5 doesn't work for rails 6.1


chuchiperriman:
Draper 4.0.5 stops working for rails 6.1.7.8. The error is: Invoice expected, got InvoiceDecorator ... which is an instance of InvoiceDecorator.

This error is raised in all method using any decorator associating the decorator a belongs_to of the parent object by example

OleksiiKachan:
rails 7.1 - same issue

PR #953, opened 2025-11-15
Release v4.0.6

seanlinsley:
This temporarily reverts #933 since that was a breaking change that shouldn't be included in a patch release, addressing #951

We can add it back after this version is released, and bump the version to 5.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Draper Changelog
 
+## 4.0.6 - 2025-11-15
+
+### Fixes
+* Revert breaking change from v4.0.5
+
 ## 4.0.5 - 2025-11-12
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


Alexander-Senko (approved):

Alexander-Senko:
> We can add it back after this version is released, and bump the version to 5.0.0

There are other changes still waiting for that.

seanlinsley:
This CI job seems to stall indefinitely: https://github.com/drapergem/draper/actions/runs/19391461282/job/55486115122?pr=953

It does have this deprecation warning:
>DEPRECATION WARNING: ActiveSupport::Configurable is deprecated without replacement, and will be removed in Rails 8.2.
> You can emulate the previous behavior with `class_attribute`.

Which seems to be coming from the unmaintained active_model_serializers gem:
```
grep -r "ActiveSupport::Configurable" $(bundle show --paths) .
actionpack-8.0.1/lib/abstract_controller/base.rb:    include ActiveSupport::Configurable
active_model_serializers-0.10.15/lib/active_model/serializer.rb:    include ActiveSupport::Configurable
```

But we're not setting `active_support.deprecation = :raise` so that shouldn't be treated as a test failure yet. So what's happening?

PR #961, opened 2026-02-13
fix: ensure turbo stream names are generated based on decorated object

### LINKED ISSUE #960: Inconsistent turbo stream names for decorated vs non-decorated objects
There's a potential very subtle regression after #928.

#928 removed the delegate of `to_global_id` to `object` in favor of it's own identification. when using `turbo_stream_from decorated_object`, Rails uses `#to_gid_param` under the hood to determine the stream name, see [here](https://github.com/hotwired/turbo-rails/blob/main/app/channels/turbo/streams/stream_name.rb#L28). the method uses `#to_global_id` to convert to `#to_param`, see [here](https://github.com/rails/globalid/blob/26544d21043a4e324db7f7a4a9c2b4a88486b388/lib/global_id/identification.rb#L46-L48). 

Because of that, a very subtle issue is introduced where it suddenly matters whether the stream name is generated from a _decorated object_, or from the _object_ itself. in a way, this is a breaking change.

## before

```
[1] [dev] (main)> Project.first.to_gid_param
"Z2lkOi8vd2VsbGZvdW5kYWkvUHJvamVjdC8x"
[2] [dev] (main)> Project.first.decorate.to_gid_param
"Z2lkOi8vd2VsbGZvdW5kYWkvUHJvamVjdC8x"
[3] [dev] (main)> Project.first.to_gid_param == Project.first.decorate.to_gid_param
true
```

## after

```
[1] [dev] (main)> Project.first.to_gid_param
"Z2lkOi8vd2VsbGZvdW5kYWkvUHJvamVjdC8x"
[2] [dev] (main)> Project.first.decorate.to_gid_param
"Z2lkOi8vd2VsbGZvdW5kYWkvUHJvamVjdERlY29yYXRvci8x"
[3] [dev] (main)> Project.first.to_gid_param == Project.first.decorate.to_gid_param
false
```

## resolving the issue

i do not know if this was an intended side effect, but when decorating objects, i was not expecting to having to track how stream names are generated. it's _very_ easy to introduce bugs because sometimes objects are decorated and sometimes just not.

i'm also not sure how to resolve the issue, given that the mentioned PR aimed to improve GlobalID compatibility in the first place.

given PRs like #929 on the other side however, it also seems like the goal is to have smooth interop with Turbo and its broadcasts as well, so I would assume this regression was not intended.

we could delegate `to_gid_param` to object for example, but i don't actually know where and how that is used, and what subtle issues that would introduce with `to_gid_param` suddenly not being `to_global_id.to_param`.

if you have thoughts on this, i'd be glad to implement a PR to resolve this issue!

markokajzer:
## Description

closes #960 via suggested fix.

#928 introduced a subtle issue where turbo stream names are generated based on the _decorated_ object, not the actual object as before.

### DIFF
--- a/spec/dummy/spec/models/post_spec.rb
+++ b/spec/dummy/spec/models/post_spec.rb
@@ -16,8 +16,14 @@ RSpec.describe Post do
   end
 
   describe '#to_sgid_param' do
+    include ActiveSupport::Testing::TimeHelpers
+
     let(:post) { Post.create }
     subject { post.to_sgid_param }
+
+    around do |example|
+      freeze_time { example.run }
+    end
 
     it { is_expected.to eq post.decorate.to_sgid_param }
   end


Alexander-Senko (approved):
The last thing is to figure out why SGIDs fail to match half of the time.

markokajzer:
it seems like there's a timing related flake. A Signed Global ID seems to include an expiration timestamp in the ms range. Example:

```
{"_rails":{"data":"gid://dummy/Post/1","exp":"2026-03-14T14:20:07.394Z","pur":"default"}}
```

To get around that, I included `ActiveSupport::Testing::TimeHelpers` for this spec (since `rails` is loaded), and used `freeze_time` around it. In the run the expiration now ends in `.000Z` for both generations.

Let me know if that works 👍

PR #962, opened 2026-02-19
CI against Ruby 4.0

chubchenko:
## Description

Ruby 4.0 has been released. This PR adds it to CI to ensure Draper works with it.


y-yagi:
@chubchenko Thanks for your contribution!

